PR #6, opened 2017-02-28
Handle error and instruct the user to generate a lockfile before audit

zmanian:
Provides an Error notification and an instruction to the user to "Run \"Cargo build\" to generate lockfile before running audit".

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -28,9 +28,9 @@ struct Package {
     version: String,
 }
 
-fn load_lockfile(filename: &str) -> Vec<Package> {
+fn load_lockfile(filename: &str) -> Result <Vec<Package>, std::io::Error> {
     let path = Path::new(filename);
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -132,7 +132,14 @@ fn main() {
         panic!("cargo-audit is intended to be invoked as a cargo subcommand");
     };
 
-    let dependencies = load_lockfile(filename);
+    let dependencies = match load_lockfile(filename){
+        Ok(v) => v,
+        Err(_) => {
```
tarcieri: `.unwrap_or_else()` might be nice here...

src/main.rs
```diff
@@ -132,7 +132,14 @@ fn main() {
...
+            notify(&mut stdout, term::color::RED,"Error",
+            "Run \"Cargo build\" to generate lockfile before running audit");
```
tarcieri: Hmm too late, but this probably should've been lower case

PR #31, opened 2018-07-23
Update `term` to 0.5 series

branan:
Term 0.4 relied on the old winapi 0.2 family of crates, which was leading to duplicate dependencies when combined with other crates that rely on the updated winapi 0.3.

This update means that `cargo-audit` now depends on only a single version of the `winapi` crate when built on Windows.


tarcieri:
Thanks!

PR #48, opened 2018-10-15
Allows macro_use that is unused on beta channel

colindean:
Beta channel builds were failing on this from the `deny(warnings)`. This enables building against all of 1.25.0, stable, and beta as of the present.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,6 +9,7 @@
 mod shell;
 
 extern crate gumdrop;
+#[allow(unused_imports)]
 #[macro_use]
 extern crate gumdrop_derive;
 extern crate isatty;


tarcieri:
Looks like an unrelated failure in one of the dependencies. I'd be fine with bumping the minimum Rust version if that resolves the issue.

colindean:
That comports with my analysis, as well. A transitive dependency now relies on a newer Rust.

The absence of `fs::read_to_string` appears to be what's blocking it from working. [It appears in 1.26.0](https://doc.rust-lang.org/std/fs/fn.read_to_string.html).

tarcieri:
I think bumping the minimum required version to 1.26 is probably for the best

PR #74, opened 2019-05-10
Update dependencies to the latest versions

chills42:
This just updates the project dependencies to the latest versions.


tarcieri:
Is there something specific which motivated this?

This seems to have added a lot of dependencies. I'd like to make a pass through at least to see what they are and if any are unnecessary.

chills42:
Just generally because versions were a bit behind and could use an update. The biggest change is due to the `term` update.

tarcieri:
This PR is out of date. My `term`-related concerns are addressed in #83, and I went ahead and did a `cargo update` as part of that change.

PR #75, opened 2019-05-19
Implement `--ignore` option.

### LINKED ISSUE #53: Add an --ignore option
Currently RustSec advisories are only as fine grained as individual crates, however there are circumstances where an advisory is not applicable, e.g. it requires a certain cargo feature to be enabled, or affects an API which a project isn't using.

While it would be nice to make advisories more fine-grained and apply to specific cargo features, or integrate with tools like `cargo-geiger` to only show advisories that actually impact a particular project, in absence of such features `cargo-audit` users presently have no mechanism to opt out of advisories which do not apply.

The carrot-and-stick answer would be for them to simply upgrade regardless, however I think adding an `--ignore` option might be a good stopgap, e.g.:

```
$ cargo audit --ignore RUSTSEC-2017-0001
```

This would allow the project to pass an audit despite having a known vulnerability.

While I see downsides to this approach, the bigger worry I'd have is losing `cargo-audit` users because of false positives and the absence of workarounds or finer-grained auditing features.

Any objections or alternatives people have in mind?

jsirois:
Also setup an integration test and cleanup clippys.

Fixes #53


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name        = "cargo-audit"
 description = "Audit Cargo.lock for crates with security vulnerabilities"
-version     = "0.6.1"
+version     = "0.7.0"
```
tarcieri: Don't bump this quite yet. I bump it in the commit where I cut a new release.

Cargo.toml
```diff
@@ -24,3 +24,6 @@ rustsec = "0.11"
 term = "0.5"
 serde_json = "1"
 
+[dev-dependencies]
+assert_cmd = "0.11"
+tempdir = "0.3.7"
```
tarcieri: `tempdir` is deprecated (use `tempfile`)

tests-data/base64_vuln/Cargo.lock
```diff
@@ -0,0 +1,25 @@
+# This file is automatically @generated by Cargo.
```
tarcieri: A couple suggestions here:

- How about putting this under `tests/support` instead of `tests-data`?
- All `cargo audit` cares about is `Cargo.lock`, so I think the rest of the app boilerplate is unnecessary.

Cargo.toml
jsirois: Reverted.

Cargo.toml
jsirois: Fixed.

tests-data/base64_vuln/Cargo.lock
jsirois: I moved to tests/support and streamlined, but left the absolute minimal Cargo.toml in place. It seemed to me we'd want this for updating Cargo.lock when new tests were added / just for provenance / proof Cargo.lock is the generated form we should expect.

Say it once more though and I'll kill Cargo.toml and dummy.rs.

tarcieri: Seems fine now

src/main.rs
```diff
@@ -297,6 +307,10 @@ fn match_vulnerability(vuln: &Vulnerability, opts: &AuditOpts) -> bool {
         }
     }
 
+    if opts.ignore.contains(&vuln.advisory.id.as_str().to_owned()) {
+        return false;
+    }
```
tarcieri: Hmm, branching on a boolean to determine a boolean return value.  It feels both unnecessary, but also like removing it wouldn't particularly improve code clarity. What do you think?

tests/integration.rs
```diff
@@ -0,0 +1,136 @@
+extern crate assert_cmd;
```
tarcieri: As it were I've been working on a sort of integration testing framework for exactly this sort of thing in the app framework I've been hoping to move this tool to. I'll accept this for now though... I can port it over!

src/main.rs
jsirois: I agree there is a bit of tension. I chose to stick with the local code style since it does have the benefit of making each filter check added read as independently affecting the result:
```
[filter1]
[filter2]
[filter3]
return true
```

Switching would be more concise but I think less clear. Now the `ignore` check works a bit different from all the rest in the priviledged final position.
```
[filter1]
[filter2]
return ![filter3]
```

tests/integration.rs
jsirois: Aha - I'll be interested in what you come up with. I had not tested mains in rust yet and `assert_cmd` was the first helper I found.

tarcieri: The approach is a bit different in that it's a framework-integrated testing approach that pulls the application in as a Rust library rather that invoking it as a subprocess, which while not providing a full from-the-terminal "integration" test does allow it to do a much better job reporting unhandled panics, with the idea that things like terminal interactions can be integration tested at the framework level.

src/main.rs
tarcieri: Seems ok

tests/integration.rs
jsirois: Nice.

dbrgn:
Any chance for a release with this feature? I need it in a project to ignore the no-fix-available ncurses vulnerability (which doesn't actually apply to that project).

tarcieri:
@dbrgn see #84

dbrgn:
@tarcieri thank you!

PR #77, opened 2019-07-01
Make the tool produce valid JSON when it detects no vulnerabilities

### LINKED ISSUE #73: --json output is invalid when there aren't vulnerabilites
Some sort of falsy output that's valid JSON like `""`, `null`, `{}`, or `[]` would be nice to make cargo-audit output easier to use as an input to other tools e.g. 

```console
» cargo audit --json > cargo-audit-output.json
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 20 security advisories (from /home/gguthe/.cargo/advisory-db)
    Scanning Cargo.lock for vulnerabilities (38 crate dependencies)
     Success No vulnerable packages found
» cat cargo-audit-output.json
» cat cargo-audit-output.json | python -m json.tool
Expecting value: line 1 column 1 (char 0)
```

I'd be happy to submit a PR for this if it's a desirable change.

jakubadamw:
Fixes #73.


jakubadamw:
Whilst getting this change ready I was facing test failures when using more than one test thread (`--test-threads` argument in `cargo test`). Turns out, `cargo audit` as such will misbehave (with `rustcsec::Repository::fetch()` errors) if there occur two concurrent first-time runs (i.e. in an environment without an advisory database repository in place). Something to keep in mind. To reduce test flakiness, I made the tests "serial", i.e. enforced that they don't get run in multiple threads.

tarcieri:
@jakubadamw FYI, I plan on moving much of the existing boilerplate over to this fairly soon:

https://docs.rs/abscissa/0.1.0/abscissa/

It includes a testing subsystem, which probably doesn't yet tackle this particular case, but could:

https://docs.rs/abscissa/0.1.0/abscissa/testing/index.html

`cargo` has a registry lock, but I think the testing subsystem could provide a `flock(2)`-style lock around executing subcommands, which would provide a sort of global mutex around CLI acceptance tests which rely on a particular directory state.

PR #79, opened 2019-07-05
Update README.md

gedigi:
added `cargo generate-lockfile` to Travis CI configuration to make sure `Cargo.lock` is generated before running `cargo audit`.


tarcieri:
Thanks!

PR #109, opened 2019-09-08
Display dependency trees with each advisory

tarcieri:
Shows a complete tree of dependencies associated with each vulnerable package in the lockfile.

This is useful for determining how vulnerable packages are being pulled into the project.

Screen Shot 2019-09-08 at 10 51 23 AM


tarcieri:
I looked at just shelling out to `cargo-tree` if it's installed, and it can certainly print trees like this, but it seems like good functionality to have built-in, and may be useful to have in the future for implementing a `cargo audit fix` feature.

PR #144, opened 2019-10-07
Print the fact that no solution is available

hbina:
Previously, if no solution is available, it will simplyprint `Solution: upgrade to <<nothing>>`.
This commit changes that and will mention the fact that no advisable upgrade is available...


tarcieri:
Nice! Thanks!

Looks like something's wrong with the GitHub Actions config and CI isn't passing the required checks. I'll take a look at fixing it.

PR #151, opened 2019-10-10
Implement --fix option

### LINKED ISSUE #23: Add a 'fix' command to automatically update dependencies.
Similar to https://docs.npmjs.com/cli/audit

Running `cargo audit fix` or a similar command would install all compatible (as in SemVer compatible, no automatic major version bumps) updates to all vulnerable packages.
This would also rewrite version constraints in `Cargo.toml`.
I realize this may be a bigger issue, I just wanted to get this idea out there.

hbina:
This option allows the program to automatically perform the recommended (if available) upgrade for the user.

Resolves #23 

Not entirely sure if the code is within reasonable idiomatic `Rust` standard.

Since, I am using `cargo-edit` for this, the way I did it is a lot simpler that what was suggested [here](https://github.com/RustSec/cargo-audit/issues/23#issuecomment-538666084). So I am not entirely sure if my PR is correct or is aligned with what @tarcieri had in mind.

I think there are niche cases that I am not handling (what happen if there is no upgrade?). In particular, this is just clumsy code on my part,

`vulnerability.versions.patched[0].to_string().as_str(),`

The suggestions laid out in Zulip is to perform some arithmetic to figure out the suitable versions to upgrade to. But I find this an unnecessary work, and I must admit I am not entirely sure if I am capable of implementing them...not without some hand holding.

Edit: This PR is also missing tests.

Edit2: Tests are failing because of dependency failures (the required [commit](6f5b85ebac2fc64aaface5242aa241bad720bc61) have not been published to crates.io yet :<

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,8 @@ petgraph = "0.4"
 rustsec = { version = "0.15.2", features = ["dependency-tree"] }
 serde = { version = "1", features = ["serde_derive"] }
 serde_json = "1"
-cargo-edit = { git = "https://github.com/killercup/cargo-edit#cargo-upgrade" }
+cargo-edit = { git = "https://github.com/killercup/cargo-edit.git" }
+
 [dev-dependencies]
 tempfile = "3"
 toml = "0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ petgraph = "0.4"
 rustsec = { version = "0.15.2", features = ["dependency-tree"] }
 serde = { version = "1", features = ["serde_derive"] }
 serde_json = "1"
-
+cargo-edit = { git = "https://github.com/killercup/cargo-edit#cargo-upgrade" }
 [dev-dependencies]
 tempfile = "3"
 toml = "0.5"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -24,7 +24,7 @@ petgraph = "0.4"
 rustsec = { version = "0.15.2", features = ["dependency-tree"] }
 serde = { version = "1", features = ["serde_derive"] }
 serde_json = "1"
-
+cargo-edit = { git = "https://github.com/killercup/cargo-edit#cargo-upgrade" }
```
tarcieri: It looks like Cargo doesn't like this dependency?

https://github.com/RustSec/cargo-audit/pull/151/checks?check_run_id=255223348

I think you can probably just do:

```
cargo-edit = { git = "https://github.com/killercup/cargo-edit.git" }
```

hbina:
Hmmm that will actually not work. If I have to guess, its because the repository actually contains 3 crates: `cargo-upgrade`, `cargo-add` and `cargo-rm`.

tarcieri:
@hbina I did a ninja edit to your branch, and the tests are now passing

tarcieri:
@hbina thank you for your contribution, but I decided to go with #157 as it seemed more complete

PR #157, opened 2019-10-22
Fix / upgrade vulnerable packages implementation

aug2uag:
Implementation for evaluation on issue #23 

`fix` added as a subcommand `CargoAuditCommand` with an update to the CLI that causes enumeration on args for match with "fix" to initiate the audit sequence with a conditional on where vulnerabilities are evaluated in presenter. The `OutputConfig` struct was changed to handle an additional boolean to define fix and upgrade state.


tarcieri:
@aug2uag thanks!

PR #175, opened 2019-12-19
Fixes #169 - Dependency on workspace crate are handled wrong

### LINKED ISSUE #169: Dependency on workspace crate are handled wrong
If you have a dependency on another crate in your workspace and it has the same name as a cargo crate, `cargo audit` resolves it to cargo one and not the actual workspace crate. 

Example Cargo.toml: 
```toml
[dependencies]
generator = { path = "../generator" }
```
`audit` looks for vulnerabilities about [generator](https://crates.io/crates/generator). So basically the field `path` is not considered here.

kegesch:
This merge should wait for [RustSec/rustsec-crate#128](https://github.com/RustSec/rustsec-crate/pull/128)

### DIFF
--- a/audit.toml.example
+++ b/audit.toml.example
@@ -29,4 +29,4 @@ arch = "x86_64" # Ignore advisories for CPU architectures other than this one
 os = "linux" # Ignore advisories for operating systems other than this one
 
 [packages]
-scope = "all" # "all", "publiccrates" or "localcrates"
+scope = "All" # "All", "Public" or "Local"
--- a/audit.toml.example
+++ b/audit.toml.example
@@ -29,4 +29,4 @@ arch = "x86_64" # Ignore advisories for CPU architectures other than this one
 os = "linux" # Ignore advisories for operating systems other than this one
 
 [packages]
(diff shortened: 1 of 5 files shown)


kegesch:
When are planning to make a new release for `rustsec-crate`? I have to wait for it right?

tarcieri:
@kegesch feel free to switch to a git dependency / add a patch directive. I'll cut a release before another release of `cargo-audit`.

audit.toml.example
```diff
@@ -27,3 +27,6 @@ show_tree = true # Show inverse dependency trees along with advisories (default:
 [target]
 arch = "x86_64" # Ignore advisories for CPU architectures other than this one
 os = "linux" # Ignore advisories for operating systems other than this one
+
+[packages]
+source = "All" # "All", "Public" or "Local"
```
tarcieri: Would prefer lowercase names for these

kegesch: Indeed.

tarcieri: I'll go ahead and merge this PR. Would appreciate followups for the above, otherwise I can take care of them before I cut final releases.

Issue #195, opened 2020-01-25
Split warning categories (i.e. have `-D` take an argument)


tarcieri:
Right now both unmaintained crates and yanked crates cause warnings, but there's no way to deny warnings for one or the other as opposed to both.

Other Rust tooling has different categories of warnings, and the `-D` option takes an argument for which ones to deny.

Ideally we could:

- `-D unmaintained` to only deny unmaintained crates
- `-D yanked` to only deny yanked crates

PR #206, opened 2020-03-09
specify deny-warning types

mmou:
Dependent on modifications in rustsec (https://github.com/RustSec/rustsec-crate/pull/156)

========

This is an initial attempt at #195 that I'm hoping for some feedback on.

I erred towards not modifying the `rustsec` crate in the approach here because I wasn't sure about the crates that depend on `rustsec`, but I think it could be cleaner to do some work there, such as:
- have the `Report` struct have something like a `WarningInfo` struct (to replace the `Vec<Warning>`)
- have the `WarningInfo` struct essentially be a hashmap of `warning::Type` to `Vec<Warning>`
- modify `Warning` and/or `warning::Type` to better support the hashmap of warning types to lists of warnings; e.g. maybe ideally `warning::Type` enum (or something like it) would not hold data

### DIFF
--- a/src/commands/audit.rs
+++ b/src/commands/audit.rs
@@ -6,7 +6,7 @@ mod fix;
 use super::CargoAuditCommand;
 use crate::{
     auditor::Auditor,
-    config::{AuditConfig, OutputFormat, WarningKind},
+    config::{AuditConfig, DenyWarningOption, OutputFormat},
     prelude::*,
 };
 use abscissa_core::{config::Override, terminal::ColorChoice, FrameworkError};
@@ -19,7 +19,7 @@ use std::{path::PathBuf, process::exit};
 use self::fix::FixCommand;
 
 /// The `cargo audit` subcommand
-#[derive(Command, Default, Debug, Options)]
+#[derive(Default, Debug, Options, Command)]
 pub struct AuditCommand {
     /// Optional subcommand (used for `cargo audit fix`)
     #[cfg(feature = "fix")]
@@ -55,7 +55,7 @@ pub struct AuditCommand {
         long = "deny-warnings",
         help = "exit with an error if any warning advisories of specified kinds are found"
     )]
-    deny_warnings: Vec<WarningKind>,
+    deny_warnings: Vec<DenyWarningOption>,
 
     /// Path to `Cargo.lock`
     #[options(
@@ -180,11 +180,11 @@ impl Override<AuditConfig> for AuditCommand {
 
         for kind in &self.deny_warnings {
             match kind {
-                WarningKind::All => {
+                DenyWarningOption::All => {
                     config.output.deny_warnings = vec![
-                        WarningKind::Other,
-                        WarningKind::Unmaintained,
-                        WarningKind::Yanked,
+                        DenyWarningOption::Other,
+                        DenyWarningOption::Unmaintained,
+                        DenyWarningOption::Yanked,
                     ]
                 }
                 k => config.output.deny_warnings.push(k.clone()),
--- a/src/commands/audit.rs
+++ b/src/commands/audit.rs
@@ -6,7 +6,7 @@ mod fix;
 use super::CargoAuditCommand;
 use crate::{
     auditor::Auditor,
-    config::{AuditConfig, OutputFormat},
+    config::{AuditConfig, OutputFormat, WarningKind},
     prelude::*,
 };
 use abscissa_core::{config::Override, terminal::ColorChoice, FrameworkError};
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -128,6 +129,47 @@ impl OutputConfig {
     }
 }
 
+/// Warning kinds
+#[derive(Copy, Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub enum WarningKind {
```
mmou: here's an attempt at options for deny-warning

src/presenter.rs
```diff
@@ -81,21 +81,63 @@ impl Presenter {
             self.print_vulnerability(vulnerability, &tree);
         }
 
+        let deny_unmaintained = self
```
mmou: Parsing `report.warnings` here instead of storing warnings by type in `Report` struct in `rustsec` crate

src/presenter.rs
```diff
@@ -81,21 +81,63 @@ impl Presenter {
...
+            .config
+            .deny_warnings
+            .contains(&WarningKind::Unmaintained);
+        let deny_yanked = self.config.deny_warnings.contains(&WarningKind::Yanked);
+
+        //
+        //  could instead do something like this (if Warning implemented Copy):
+        //
+        //  let (denied, others): (Vec<Warning>, Vec<Warning>) =
+        //     report.warnings.iter().partition(|&w| match w.kind {
```
mmou: Couldn't use `partition` because `Warning` doesn't implement `Copy`...

src/presenter.rs
```diff
@@ -81,21 +81,63 @@ impl Presenter {
...
-            let warning_word = if report.warnings.len() != 1 {
+            let warning_word = if num_denied != 1 {
                 "warnings"
             } else {
                 "warning"
             };
 
-            if self.config.deny_warnings {
-                status_err!("{} {} found", report.warnings.len(), warning_word);
+            status_err!("{} denied {} found", num_denied, warning_word);
```
mmou: Some decisions to be made about how to display this...

src/presenter.rs
```diff
@@ -106,7 +148,7 @@ impl Presenter {
 
             let msg = "this copy of cargo-audit has known advisories!";
 
-            if self.config.deny_warnings {
+            if self.config.deny_warnings.contains(&WarningKind::Other) {
```
mmou: Currently considering self advisories to be `WarningKind::Other`

tarcieri:
Thanks for taking a look at this!

I would suggest modifying the `rustsec` crate, and not worry too much about the downstream consumers, particularly for warnings. The only other notable crate that depends on `rustsec` directly is `cargo-deny`, and I don't think it consumes this information. Regardless, the API is still unstable so it's ok to change.

> have the WarningInfo struct essentially be a hashmap of warning::Type to Vec<Warning>

I think this would be the best approach.

src/presenter.rs
tarcieri: Generally I've looked to `rustc` and `clippy` for ideas and conventions for this sort of thing.

They promote `warning:` -> `error:` in this case, and add a `note:` saying that deny warnings has been enabled.

Cargo.toml
```diff
@@ -19,7 +19,7 @@ abscissa_core = "0.5.2"
 gumdrop = "0.7"
 home = "0.5"
 lazy_static = "1"
-rustsec = { version = "0.18", features = ["dependency-tree"] }
+rustsec = { git = "https://github.com/mmou/rustsec-crate", branch = "mmou/issue-195-warnings", features = ["dependency-tree"] }
```
mmou: temporary

src/commands/audit.rs
```diff
@@ -213,7 +225,7 @@ impl Runnable for AuditCommand {
             exit(0);
         }
 
-        let lockfile_path = self.file.as_ref().map(PathBuf::as_path);
+        let lockfile_path = self.file.as_deref();
```
mmou: clippy told me to use as_deref?

Cargo.toml
mmou: waiting for next rustsec release.

tarcieri:
Note: #221 includes changes which conflict with / duplicate some of this PR.

However, I think it should make it easier to implement, since it bumps the `rustsec` crate to the latest release.

mmou:
> Note: #221 includes changes which conflict with / duplicate some of this PR.
> 
> However, I think it should make it easier to implement, since it bumps the `rustsec` crate to the latest release.

Thanks for the FYI! I updated the branch, rebased off of master. 

I need your help fixing the Self-Audit CI step ` cargo run -- audit --deny-warnings --ignore RUSTSEC-2019-0031`; should be updated to: ` cargo run -- audit --deny-warnings all --ignore RUSTSEC-2019-0031`.

Issue #231, opened 2020-05-29
README error: --dry-run requires an argument


xobs:
The README says that you can do a dry run by appending `--dry-run`  to the command. However, this does not actually work:

```sh
$ cargo audit fix --dry-run
error: missing argument to option `--dry-run`

cargo-audit 0.12.0
Tony Arcieri <bascule@gmail.com>
Audit Cargo.lock for crates with security vulnerabilities

FLAGS:
    -c, --config CONFIG       path to configuration file
    -h, --help                print help message
    -v, --verbose             be verbose
$
```

It appears as though the `dry-run` option requires a value, and `true` can be passed.

The README should give the command as: `cargo audit fix --dry-run true`

tarcieri:
This is an oversight. I'll try to fix the flag so as not to require an argument.

tarcieri:
This was fixed in #245.

PR #245, opened 2020-07-10
fixed --dry-run requires argument #231

vishalsodani:


tarcieri:
Thanks! The test failures look unrelated. I'll try to address those in a separate PR.

PR #255, opened 2020-09-21
pin smol_str to 0.1.16 to support Rust 1.44 or older

### LINKED ISSUE #254: cargo audit does not work on Rust 1.44 anymore
While running `cargo install cargo-audit --force` with Rust 1.44 we are getting following [error in smol_str](https://github.com/rust-analyzer/smol_str/issues/25):

```
error[E0658]: `while` is not allowed in a `const fn`
  --> C:\Users\serg\.cargo\registry\src\github.com-1ecc6299db9ec823\smol_str-0.1.17\src\lib.rs:58:9
   |
58 | /         while i < text.len() {
59 | |             buf[i] = text.as_bytes()[i];
60 | |             i += 1
61 | |         }
   | |_________^
   |
   = note: see issue #52000 <https://github.com/rust-lang/rust/issues/52000> for more information
```

`cargo-audit` uses `crates-index` which uses `smol_str` and now "cargo audit requires Rust 1.40 or later." claim does not hold.

our CI runs cargo audit and must keep running on Rust 1.44 due to performance issues with 1.45-1.46.

would it be possible to pin smol_str in Cargo.lock to something that works with older compiler versions?

galich:
`smol_str` crate is used in `cargo-audit` through `crates-index`. `smol_str` MSRV is current stable version of Rust, while `cargo-audit` MSRV is 1.40. This PR pins `smol_str` version to one that works with older Rust versions.

fixes #254

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,6 +198,7 @@ dependencies = [
  "rustsec",
  "serde",
  "serde_json",
+ "smol_str",
  "tempfile",
  "thiserror",
  "toml",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ thiserror = "1"
 once_cell = "1.3"
 tempfile = "3"
 toml = "0.5"
+smol_str = "=0.1.16" # smol_str 0.1.17 is incompatible with Rust 1.44 or lower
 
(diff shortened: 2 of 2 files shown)


tarcieri:
Note: I've opened #256 to address the self-audit failure

PR #268, opened 2020-10-19
Various presenter improvements

bossmc:
* Commonize printing Metadata
* Remove some unneeded `unwrap` calls
* Print all outcomes even if we're going to fail
* Remove redundant output

I also noticed that version 0.12.1 seemed to be tagged at the wrong place in the git history (so some of the "--deny-warnings" logic is missing, and the synopsis count of warnings is wrong).


tarcieri:
The self-audit failure is unrelated (see #269 

> I also noticed that version 0.12.1 seemed to be tagged at the wrong place in the git history (so some of the "--deny-warnings" logic is missing, and the synopsis count of warnings is wrong).

v0.12.1 is effectively a backported fix which was applied to the v0.12.0 release, then merged back into `master`, in order to maintain the MSRV after downstream dependencies introduced MSRV breaking changes.

The release didn't contain any of the subsequent changes presently on `master` (deliberately).

bossmc:
I've rebased to pick up that change - so checks should pass now.

Thanks for the info about 0.12.1.  Coincidentally, I've added a `rust-toolchain` file which rustup will read and invoke the MSRV automagically, so hopefully preserving the minimal rust version will be easier in future too.

tarcieri:
Thanks! These look like nice improvements.

PR #298, opened 2020-12-29
Update README.md

SergeyKolchanov:
cargo-audit now requires rustc 1.44 because of dependency on semver-parser, which now requires rustc 1.44. For details, see:
https://github.com/steveklabnik/semver/issues/218


tarcieri:
Thanks!

PR #315, opened 2021-02-24
When running in no-fetch mode, allow accessing a non-git repo.

Firstyear:
Many offline use cases when they request --no-fetch are for offline
systems that may have their advisory db from an external source such
as an internal packaged version. In these cases when we are flagged
as no fetch, we should open the database with "open" instead of
"load from repo" to support operating without git metadata.

This also sets that when in no-fetch crates.io should not be updated
for the yanked list.

Fixes https://github.com/RustSec/cargo-audit/issues/313

### DIFF
--- a/src/auditor.rs
+++ b/src/auditor.rs
@@ -44,12 +44,12 @@ impl Auditor {
             .cloned()
             .unwrap_or_else(rustsec::GitRepository::default_path);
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Thanks! Self-audit failure is unrelated

Issue #66, opened 2019-01-05
Pre-built distribution / releases


passcod:
It would be nice to have pre-built binaries of the tool available here. This is mostly for CI usage as building the tool in CI inflates build times. This is less an issue with caching, but is still a concern.

tarcieri:
One option for this I had suggested on a [rust-internals thread about upstreaming RustSec](https://internals.rust-lang.org/t/pre-rfc-reviving-security-advisories-in-crates-io-rfc-pr-1752/9017/7) was to distribute `cargo audit` as a rustup component.

tarcieri:
I consider binary releases to be a large responsibility and something that should be done right or not done at all, especially given the security-critical nature of this tool (that said, I do spend a lot of time doing build / release automation for private/enterprise Rust projects).

Here is a quick list of questions about what that might look like:

1. How will binary releases be built? If there's a community build serve / release automation (e.g. @bors) available, I'd be ok with using something like that, but AFAIK there is not (short of releasing as a rustup component). I am **not** ok with "here's a script to build it on your personal computer", and would be strongly in favor of using build automation for releases. Travis CI is a possibility, although my personal preference for something like this is GCP Cloud Build.
2. Which operating systems should receive binary releases? My experience is almost entirely Linux-centric, so unless someone is volunteering to do the work for other operating systems that sounds like the logical starting place.
3. For Linux...
    a. What artifacts will be delivered? Are we producing a static binary linked to musl, or distro-specific packages?
    b. How will release artifacts be delivered? Are we making a tarball of aforementioned static library, or distro-specific packages which can be apt/yum/dnf-installed? All of the above?
    c. How will releases be authenticated? Are we GPG signing that tarball, or using distro-specific GPG integrations to sign distro-specific packages?

If there are satisfactory answers to enough of these questions, and people doing the work, I can consider independent binary releases...

...but personally I *really* don't want to do any of this work, because I'd rather invest the time trying to upstream this tool, which would solve the same problem by making it available as a rustup component.

tarcieri:
tl;dr: of the above: this is something I'm too paranoid to delegate, with too much maintenance burden for me to operate pretty much regardless of the solution, and too much margin for error in a security-critical tool for what feels like a convenience win.

I think the path forward here is upstreaming the project and producing a rustup component using the existing rustup build pipeline.

tarcieri:
I'll throw one option out here which might make sense given my personal circumstances (but may not be what people are generally after here):

[...]

PR #332, opened 2021-03-08
Build weekly Docker images

dbrgn:
Build a docker image based on `rust:latest`, which installs latest cargo-audit from crates.io.

Note: This will not create versioned releases, the `latest` tag will always contain the latest crates.io release! This is ideal for use in CI, where you want to keep up to date. Always tagging the currently built image with the version would be easy, but it means that only the latest released version would get image updates, which is not ideal. A setup where multiple versions are regularly re-built, would be more complex and is out of scope for now.

Images are automatically built and pushed every Sunday at 13:37 UTC, and when pushing to `main`. The job does not run on forks (because the push would fail).

I think this should be ready now, but you'll still need to set up the `DOCKER_USERNAME` and `DOCKER_PASSWORD` secret variables. Make sure to create a dummy account for uploading.

Refs #66.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,36 @@
+name: Publish Docker Image
+
+# Build on push to main and every week (sunday at 13:37)
+on:
+  push:
+    branches: [main]
+  schedule:
+    - cron: '37 13 * * 0'
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: -D warnings
+  DOCKER_IMAGE: rustsec/cargo-audit
+
+jobs:
+  # Build a docker image based on rust:latest, which installs latest cargo-audit from crates.io.
+  #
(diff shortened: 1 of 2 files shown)


dbrgn:
Note: If you want to test this on a branch, create a new branch from my fork, add it to the `push.branches` list in `docker-image.yml` and then push the resulting commit to GitHub.

tarcieri:
Perhaps we should discuss on #66, but I have a very specific view of the deliverables of that PR, and this doesn't really seem to be in line with what I have in mind at all.

I'll post a follow-up comment there.

tarcieri:
Closing in favor of #337

dbrgn:
> Closing in favor of #337

👍🏻 

FYI, in case someone else needs this, I'm regularly publishing Docker images of the latest released version of cargo-audit to https://hub.docker.com/r/dbrgn/cargo-audit.

PR #343, opened 2021-04-29
Vendor OpenSSL for arm and musl builds

passcod:
Ref #337


passcod:
I've based the branch off the commit prior to the last versioning commit if you want to redo the release, otherwise... until next time.

PR #357, opened 2021-05-09
Add some tier 3 targets

messense:
Too many of them...😱


tarcieri:
Looks like it needs a rustfmt

tarcieri:
Thank you!

Issue #349, opened 2020-09-24
Pre-release handling regression


tarcieri:
RustSec/rustsec-crate#217 updated the `semver` crate to v0.10, and with it ripped out a lot of vendored code from `semver` v0.9 (as it were, RustSec/rustsec-crate#244 has since bumped `semver` to v0.11)

This likely caused a regression of how prereleases are handled. See:

- https://github.com/RustSec/cargo-audit/issues/17
- https://github.com/RustSec/cargo-audit/issues/30
- https://github.com/steveklabnik/semver/issues/172

Shnatsel:
I am working on fixing this. A solution for this arrives as a side effect of the OSV integration work, see #366

Shnatsel:
The fix has shipped in `rustsec` v0.24 and cargo-audit v0.15

PR #366, opened 2021-05-18
OSV export

Shnatsel:
This PR implements export from RustSec TOML format to the JSON-based format defined by https://github.com/google/osv

The implementation is feature-complete. The generated JSON has been successfully imported by OSV. See [here](https://github.com/Shnatsel/advisory-db/tree/osv) for what the exported data looks like. 

**Summary of the changes:** 

 - Implemented conversion from Cargo-like selectors such as `"> 2.0.0", "<= 2.5"` or `"^1.5"` to version ranges, see `OsvRange` struct. This is required for OSV export.
 - Migrated our version matching logic to using OSV version ranges. This fixes [a long-standing bug around handling of pre-release versions](https://github.com/RustSec/rustsec/issues/349).
 - Added conversion from RustSec `Advisory` to data structures isomorphic to OSV JSON, which can be serialized with `serde_json`.
 - Implemented retrieval of file modification time from git history, which is required for OSV's `modified` field; see ` GitModificationTimes` struct.
 - Added `withdrawn` field to RustSec advisory format, which must be set to a date. This is required for exporting to OSV. Updated the linter to require this field to be present if `yanked = true`. This is a breaking change under [semantic versioning](https://semver.org/) and caused a version bump.
 - Now that we're breaking the API anyway, changed the `Versions` struct to validate range consistency on creation, so we don't have to check for inconsistent ranges in every single function dealing with them.
 - Added `rustsec-admin osv` CLI command for exporting the RustSec TOML to OSV JSON.

**Remaining merge blockers:** 

 - None

**Remaining blockers for deploying continuous export to OSV:**

 - [ ] Add a CI job calling `rustsec-admin osv` - implemented [here](https://github.com/Shnatsel/advisory-db/blob/main/.github/workflows/export-osv.yml)
 - [ ] Add the `withdrawn` field to the production advisory DB: https://github.com/RustSec/advisory-db/pull/942


Shnatsel:
Let me know if you'd prefer to keep the PR in draft or close it to avoid the cost of running CI on every commit.

tarcieri:
It's fine to keep it open

Shnatsel:
This is now feature-complete. The list of remaining work items can be found here: https://github.com/Shnatsel/rustsec/blob/osv/OSV-TODO.md

rustsec/src/repository/git/modification_time.rs
```diff
@@ -0,0 +1,56 @@
...
+        let mut mtimes: HashMap<PathBuf, Time> = HashMap::new();
+        let repo = git2::Repository::open(repo.path())?;
+        let mut revwalk = repo.revwalk()?;
+        revwalk.set_sorting(git2::Sort::TIME)?;
+        revwalk.push_head()?;
+        for commit_id in revwalk {
+            let commit_id = commit_id?;
+            let commit = repo.find_commit(commit_id)?;
+            // Ignore merge commits (2+ parents) because that's what 'git whatchanged' does.
+            // Ignore commit with 0 parents (initial commit) because there's nothing to diff against
```
Byron, line 32: Would that also mean that if a file is touched only in the initial commit it would be missing entirely? Could that be relevant to the generated advisory output in the form of a missing advisory?
Asked differently, would it hurt to diff against an empty tree (`None`) in that case to also fetch these paths just in case?

Shnatsel: In the general case you're probably right, but this is not relevant for the advisory database which was initialized with a README, and which has been touched since.

admin/src/osv_export.rs
```diff
@@ -0,0 +1,96 @@
...
+                    if let Some(kind) = &advisory.metadata.informational {
+                        match kind {
+                            // If not `Unmaintained` or `Unsound` or `Notice`, don't export it to OSV
+                            // to make the output format stable.
+                            // Adding new types should be accompanied by a version bump.
+                            Informational::Unmaintained => (),
+                            Informational::Unsound => (),
+                            Informational::Notice => (),
+                            _ => continue,
+                        }
```
tarcieri, line 66: I'm kind of surprised `clippy` doesn't suggest grouping these in some way... either a `match` with them all `|`'d, a `matches!` macro, or an array with `contains?`.

It's fine I guess, it just seems a bit more verbose than it needs to be.

rustsec/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name        = "rustsec"
 description = "Client library for the RustSec security advisory database"
-version     = "0.23.3" # Also update html_root_url in lib.rs when bumping this
+version     = "0.24.0" # Also update html_root_url in lib.rs when bumping this
```
tarcieri, line 4: Usually I reserve these sort of bumps for a release commit... how about:

```suggestion
version     = "0.24.0-pre" # Also update html_root_url in lib.rs when bumping this
```

rustsec/Cargo.toml
```diff
@@ -44,6 +45,7 @@ fix = ["cargo-edit"]
 git = ["crates-index", "git2", "home", "humantime", "humantime-serde"]
 dependency-tree = ["cargo-lock/dependency-tree"]
 vendored-openssl = ["git2/vendored-openssl"]
+osv-I-know-this-is-unstable = ["git", "chrono"]
```
tarcieri, line 48: Perhaps simply `osv` suffices here, and you can leave a comment regarding stability?

tarcieri (approved):
A few small nits and minor suggestions, otherwise this seems fine

admin/src/osv_export.rs
Shnatsel: I tried using `|`, but `rustfmt` makes it look weird.

I took this over more complicated constructs because this is simple and stupid, and so easy to understand. I try to avoid clever code in order to make code easier to maintain, and immediately understandable even to someone not intimately familiar with Rust.

rustsec/Cargo.toml
Shnatsel: The format itself is not stabilized. I explicitly wanted to mark anything exporting to not-yet-finalized format as unstable, because it's not really exporting to OSV, it's exporting to a pre-release of OSV.

Also, renaming it to `osv` later will make it clear if e.g. some dependent code has been updated to the stable version or not. Not just the stable version of the code, but also the format.

rustsec/Cargo.toml
Shnatsel: Renamed to `0.24.0-beta.1`

rustsec/src/repository/git/modification_time.rs
Shnatsel: I've opened https://github.com/rustsec/rustsec/pull/440 to address this. Thanks again!

PR #368, opened 2021-05-20
Add new exit status for errors

tweksteen:
Use a new exit status (2) when an error is triggered (e.g., when finding or parsing the lockfile). Update the audit() and generate() definitions to propagate the error.

`crate::error` does not provide a convenient constructor. It is also not used elsewhere. Instead, rely on `rustsec::error` to build the error.


tarcieri:
Seems like a good idea. Thanks.

PR #371, opened 2021-05-21
Fix CI breakage & migrate CI to ureq

Shnatsel:
Fixes CI breakage due to https://github.com/theduke/crates_io_api/issues/30 and prevents similar breakage in the future.

Migrates from `crates_io_api` crate to plain HTTP requests with `ureq`. This slightly reduces the supply chain (-2 teams, -5 publishers) and dramatically reduces the amount of unsafe code in the dependencies (from 22k to 12k expressions).

`rustls` is used for fetching data from crates.io. The only downside of it is portability to exotic CPU architectures, but since this is only meant to run on CI it doesn't matter.

However, we still use openssl for fetching data from git because libgit2 depends on it.

Currently rustls is configured to use bundled certificates (`webpki-roots`), let me know if you want the host certificates to used instead to match what OpenSSL does.


admin/src/linter.rs
```diff
@@ -134,4 +134,29 @@ impl Linter {
...
+
+        #[derive(serde::Deserialize)]
+        struct CrateInfo {
+            name: String, // there are more fields, but this is the only one we need
+        }
+
+        let url = format!("https://crates.io/api/v1/crates/{}", name);
+        let response: CrateResponse = self.http_client.get(&url).call()?.into_json()?;
+        // FIXME: I've ported this equality check from legacy code
+        // and I have no idea what it does or why it's needed
```
alex, line 159: It has to do with name normalization. A request for https://crates.io/api/v1/crates/serde-json will return `serde_json` data an we want to catch use a non-canonical name.

alex (approved):

lnicola:
Does this need authentication and/or rate limiting?

I think another approach might be to update the local cargo index and query that instead.

tarcieri:
> I think another approach might be to update the local cargo index and query that instead (modulo normalization).

The problem is we're trying to query the *package* name to ensure it's correct, and I believe the index only contains the *crate* name post-normalization.

I could be mistaken though.

Shnatsel:
Authorization: AFAIK setting a custom user agent is sufficient.

Rate-limiting: it's not clear. The API access policy only ever talks about crawling the entire API, which is not happening here. Given that this is used rarely and in short bursts, it's probably fine.

I've tried to get a clarification out of the crates.io team before for https://github.com/rust-secure-code/cargo-supply-chain, and so far I have failed. You can also look at that project for alternative approaches, which have their own downsides...

Shnatsel:
Huh, using the crates.io index should be sufficient: https://github.com/rust-lang/crates.io-index/tree/master/se/rd - this has both _ and - symbols in the names, they don't appear to be mangled.

A good thing for a follow-up PR then!

tarcieri:
We use the `crates-index` crate elsewhere in the `rustsec` crate to query the index. If the information is actually in there, that seems fine.

Shnatsel:
I'll see what I can do with `crates-index` in a follow-up PR.

Shnatsel:
I've opened #372 to use crates.io index instead. Thanks for the suggestion @lnicola !

PR #372, opened 2021-05-21
Use crates index instead of crates.io api

Shnatsel:
Well, that was easy.


tarcieri (approved):

Shnatsel:
After this change the entire advisory db is linted in 130ms with a warm cache, 3.7s with a cold cache on an HDD.

Not sure how impactful fetching crates.io index will be on CI execution time, however.

lnicola:
In rust-analyzer we cache the cargo index IIRC.

admin/src/linter.rs
```diff
@@ -34,15 +34,15 @@ impl Linter {
     /// Create a new linter for the database at the given path
     pub fn new(repo_path: impl Into<PathBuf>) -> Result<Self, Error> {
         let repo_path = repo_path.into();
-        let http_client = ureq::AgentBuilder::new()
-            .user_agent("RustSec advisory database linter")
-            .timeout(Duration::from_secs(20))
-            .build();
+        let crates_index = crates_index::Index::new_cargo_default();
+        if !crates_index.exists() {
+            crates_index.retrieve()?;
```
lnicola, line 39: Is there a way to update it in case it exists, but it's stale?

Shnatsel: Good catch! Turns out the crate examples are misleading. Fixed!

lnicola (approved):

alex (approved):

PR #390, opened 2021-06-29
Update atom_syndication to 0.10

amousset:


amousset:
Replaces https://github.com/RustSec/rustsec/pull/385

PR #393, opened 2021-07-02
Do not lint year in CVE IDs

Qwaz:
The year part of a CVE ID may be different from the reported year as discovered in RustSec/advisory-db#946. This PR softens the linting for CVE IDs as suggested in the [comment](https://github.com/RustSec/advisory-db/pull/946#issuecomment-872698576).


tarcieri:
Thanks. We'll need to bump the `rustsec` crate and `rustsec-admin` to make this work.

tarcieri:
Released in #394 but still needs a bump in `rustsec-admin` (at least on the advisory DB itself)

Issue #398, opened 2021-07-06
CVSS score for no-impact vectors computation is incorrect


ShereKahn:
When computing CVSS scores for no-impact base vectors (that is vectors where C, I, and A all are set to 'N'), the expected score should be 0.0, but the score is computed incorrectly.

E.g. https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:N/I:N/A:N and https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:N 

Both are 0.0 according to FIRST, CVSS lib is giving scores in the Medium range.

ShereKahn:
opened pull request #399 as proposed fix

ShereKahn:
closed by #399

PR #399, opened 2021-07-06
Proposed fix for #398 

ShereKahn:
'no impact' => all of C, I, A are set to N

Two changes were needed:
1) the call to .abs() lead to a non negative  iss_scoped score 
in the problematic case, removed that call

2) the iss_scoped needs to be inferior or equal to 0   for the
resulting score to be set to 0,
else the exploitability gets factored in for the final score

This is  my first ever pull request, Sorry if I missed something! 
Thanks


tarcieri:
Looks good. Thank you!

PR #402, opened 2021-07-09
Support ~ and = operators in version specification

Shnatsel:
* Add support for `~` operator in version requirements, add tests
* Add support for `=` operator in version requirements, add tests
* Make `^` operator handling spec-compliant, add tests. The spec discrepancy doesn't seem to affect the current DB data.

After this is merged, if we use `~` requirements in the DB, that will work on both very old cargo-audit versions and very new ones, but version 0.15.0 exactly will be broken (it's the only released version with rustsec v0.24 before these changes)

### DIFF
--- a/rustsec/src/osv/unaffected_range.rs
+++ b/rustsec/src/osv/unaffected_range.rs
@@ -200,6 +200,23 @@ impl TryFrom<&semver::VersionReq> for UnaffectedRange {
                     start = Bound::Inclusive(start_version);
                     end = Bound::Exclusive(end_version);
                 }
+                Op::Tilde => {
+                    if input.comparators.len() != 1 {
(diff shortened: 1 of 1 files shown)


Shnatsel:
Following this the only unsupported operator is `*`

tarcieri:
We could potentially yank v0.15.0. People who are on it already are probably updating fairly frequently.

But it would probably still be good to give it some time to bake after cutting another release before we actually tried to use these.

tarcieri (approved):

Shnatsel:
> We could potentially yank v0.15.0. People who are on it already are probably updating fairly frequently.

That was the case until yesterday, when we broke v0.12 users :stuck_out_tongue_closed_eyes:

PR #421, opened 2021-08-27
Support OSV v0.8

Shnatsel:
Mirror the changes done in https://github.com/ossf/osv-schema/pull/1

Marking this as draft because OSV upstream is not yet prepared to pull in v0.8, but I believe the code on our side is complete.

I have raised some concerns about the v0.8 changes privately, but they mostly concern parsing and various edge cases we don't exercise, so future revisions are unlikely to affect us.

~~Curiously, the new structure lets us also encode the version ranges we've put on affected functions. I'm not sure it's a good idea to keep those around - but now if we want to, we can!~~

### DIFF
--- a/rustsec/src/osv/osv_advisory.rs
+++ b/rustsec/src/osv/osv_advisory.rs
@@ -2,7 +2,7 @@ use chrono::{DateTime, NaiveDate, NaiveDateTime, NaiveTime, Utc};
 use serde::Serialize;
 use url::Url;
 
-use super::{ranges_for_advisory, OsvRange};
+use super::ranges_for_advisory;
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
This is now ready for review.

We'll need to switch over the export to `osv-experimental-v0.8` branch and don't overwrite -v0.7. OSV is not ready to read v0.8 just yet, but should pick it up in a few days.

Shnatsel:
OSV is now ready to ingest v0.8 format. I'm going to merge this. After that we'll need new releases of `rustsec` and `rustsec-admin` crates, and finally switch CI over to the new version.

PR #436, opened 2021-09-12
Parse --color=auto correctly

ajeetdsouza:
Currently, `cargo audit --color=auto` (which the `--help` says is the default option) causes a panic. This PR fixes that.


tarcieri:
Thank you

PR #442, opened 2021-09-28
README: Add packaging status for cargo-audit

kpcyrd:
Currently looks like this:


tarcieri:
Oh wow, neat!

tarcieri:
Hmm, with the `align = "right"` this ended up looking a little weird in the README:

Screen Shot 2021-09-28 at 7 21 47 AM

kpcyrd:
hmm, indeed! The install section could be padded with example install instructions for various distros so it becomes longer than the image. I could add a few if you're interested.

PR #445, opened 2021-10-04
ci: extract version

efx:
This may relate to why the action failed. The prior index extracted `cargo-audit` instead of the version. Relates to #66


efx:
@tarcieri @passcod FYI. I have some time today to push this over the finish line.

PR #447, opened 2021-10-04
ci: fix release action

efx:
The release action was not working because it expected the build step's
`target/` to be in the working directory.

It actually was in the parent directory.

actions-rs/cargo does not support the `working-directory` option which
contributed to the path confusion.

Relates to #66.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,7 +100,7 @@ jobs:
           set -eux
           ext=""
           [[ "${{ matrix.name }}" == windows-* ]] && ext=".exe"
-          bin="target/${{ matrix.target }}/release/cargo-audit${ext}"
+          bin="../target/${{ matrix.target }}/release/cargo-audit${ext}"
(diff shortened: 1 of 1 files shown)


efx:
I've been integration testing the basic structure here:
https://github.com/efx/actions/actions

I don't know if the `release.yml` uploading was working as expected as I do not see any target specific assets in that repository's [releases UI](https://github.com/efx/actions/releases). I'm not as familiar with GitHub releases so may be missing something. I'll take a gander sometime tomorrow.

efx:
Ha, actually it looks like we cannot glob for the archived files so I'll need to push up another fix tomorrow!
https://github.com/efx/actions/runs/3796477599?check_suite_focus=true#step:10:11

tarcieri:
> actions-rs/cargo does not support the working-directory option which contributed to the path confusion.

FWIW lately I've switched to only using actions-rs for the Rust environment setup, and using `run: cargo...` to actually invoke cargo

PR #452, opened 2021-10-18
Add information about cvss score and metrics

amousset:
* Add severity tag in advisories lists
* Add score and severity in advisory details
* Add base metrics in advisory details
* Make CVSS vector a link to nvd calculator

details
list


tarcieri:
Wow, awesome! 👏

PR #468, opened 2021-11-11
Unpin thiserror and smol_str

jsgf:
Pinning makes the dependency graph brittle and hard to integrate with
other crates and their dependencies.

As far as MSRV goes, `thiserror` seems like a non-issue since it has always had an MSRV of 1.31.
 `smol_str` is no longer needed so can be removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1909,7 +1909,6 @@ dependencies = [
  "platforms",
  "semver 1.0.4",
  "serde",
- "smol_str",
  "tempfile",
  "thiserror",
  "toml",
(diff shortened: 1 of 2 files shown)


alex:
The `smol_str` pin is intentional to keep the MSRV.

tarcieri:
Notably `smol_str` was pinned because we received widespread reports of MSRV breakages from downstream users and are trying to avoid that from happening again.

Their MSRV policy is:

> Minimal Supported Rust Version: latest stable.
> Bumping MSRV is not considered a semver-breaking change.

...so we have no recourse but to pin if we would like to provide any kind of MSRV guarantees.

tarcieri:
That said, pinning with `=` is overly restrictive. It could be pinned with `<=` instead.

jsgf:
The problem is that I'm trying to combine rustsec with another package which requires a higher smol_str. Maybe the alternative is to not use smol_str if it has an incompatible MSRV policy.

tarcieri:
We only had it pinned because it was a transitive dependency, *however*...

Checking again it looks like it's no longer a transitive dependency, so it appears it can be safely removed 🎉

jsgf:
Oh cool - do you want me to respin this PR or will you remove it?

tarcieri:
Go ahead and remove it and we can merge this

alex (approved):

tarcieri:
For reference the crate that depended on `smol_str` was `crates-index`, however it appears they migrated to `smartstring`.

I'm not sure what the MSRV policy is there, but regardless, this PR seems fine to merge.

PR #471, opened 2021-11-11
rustsec: make `advisory::id::Kind` lowercase

tarcieri:
Fixes the newly-introduced `clippy::upper_case_acronyms` lint by following standard Rust capitalization conventions, changing the all upper case names to ones which capitalize only the first letter, e.g.

- `advisory::id::Kind::RUSTSEC` => `advisory::id::Kind::RustSec`
- `advisory::id::Kind::CVE` => `advisory::id::Kind::CVE`
...

### DIFF
--- a/admin/src/assigner.rs
+++ b/admin/src/assigner.rs
@@ -58,7 +58,7 @@ pub fn assign_ids(repo_path: &Path, output_mode: OutputMode) {
         let id = metadata.id;
         let year = metadata.date.year();
 
-        if let Kind::RUSTSEC = id.kind() {
+        if let Kind::RustSec = id.kind() {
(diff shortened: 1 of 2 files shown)


Shnatsel:
This is a public API break, since this is exposed in the public API: https://docs.rs/rustsec/0.24.3/rustsec/advisory/id/enum.Kind.html#variant.GHSA

Is it really worth the trouble?

tarcieri:
Yes, I'm bumping MSRV anyway, so it will be a breaking change.

There are other public API cleanups I'd also like to do.

This isn't a 1.0 crate yet. We're allowed to make breaking changes.

PR #472, opened 2021-11-11
rustsec: fix `OsvReferenceKind` cases

tarcieri:
Like #471, this renames the variants of `OsvReferenceKind` to have capitalized first letters with the rest of the letters lower case, following Rust naming idioms now enforced by the `clippy::upper_case_acronyms` lint.

cc @Shnatsel


Shnatsel (changes requested):
The OSV format specifies these as uppercase. So if you want to rename them, you'd also have to create a custom `Serialize` impl that converts the name to uppercase.

Other than that, is a public API breaking change for cosmetic reasons. I don't think it's a good idea; I suggest simply suppressing the lint.

tarcieri:
Okay!

PR #473, opened 2021-11-11
platforms: follow `upper_case_acronyms` conventions

tarcieri:
Changes the `target::Arch` and `target::Env` enums to have variants which pass the `clippy::upper_case_acronyms` lint, replacing all upper case letters with the standard Rust enum capitalization.

### DIFF
--- a/platforms/Cargo.toml
+++ b/platforms/Cargo.toml
@@ -4,7 +4,7 @@ description = """
 Rust platform registry with information about valid Rust platforms (target
 triple, target_arch, target_os) sourced from Rust Forge.
 """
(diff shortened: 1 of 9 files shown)


platforms/Cargo.toml
```diff
@@ -4,7 +4,7 @@ description = """
 Rust platform registry with information about valid Rust platforms (target
 triple, target_arch, target_os) sourced from Rust Forge.
 """
-version    = "1.1.0" # Also update html_root_url in lib.rs when bumping this
+version    = "2.0.0-pre" # Also update html_root_url in lib.rs when bumping this
```
tarcieri, line 7: Note: there are already other breaking changes that haven't been released. I'm not just bumping it for this.

PR #481, opened 2021-11-12
Update atom_syndication to 0.11

amousset:
We must not escape the html as it is done automatically by atom_syndication now.


tarcieri:
Thank you!

Issue #408, opened 2021-07-29
`cargo audit` incorrectly checks packages in non-crates.io registries


timwalls:
Some of my packages are in a private registry.  I've just discovered that `cargo audit` doesn't realise this and uses the vulnerabilities of public packages if the names happen to match.

For example, I have a package `serial` that is now on version 0.3.1:
```
[package]
name = "serial"
version = "0.3.1"
authors = ["Tim Walls <tim.walls@snowgoons.com>"]
edition = "2018"

# For publishing to our private repository
publish = ["snowgoons-crates"]
```

I have applications that reference this package like this:
```
serial = { version = "0.3", registry = "snowgoons-crates" }
```

These mysteriously started failing their audits when my serial package hit 0.3.1:

```
Crate:         serial
Version:       0.3.1
Warning:       yanked
Dependency tree:
serial 0.3.1
├── printer-svc 0.8.0
└── cash-svc 0.7.9

error: 1 denied warning found!
```

I just realised this is because there is a package of the same name in the crates.io repository, which happened to have its 0.3.1 version yanked: https://crates.io/crates/serial/versions

### TL;DR version
`cargo audit` should recognise when a package is imported from a registry other than `crates.io`.

tarcieri:
Thought we had some other open issues about this but apparently not.

The fix should be fairly straightforward.

timwalls:
Yep, I did have a quick check first and didn't find anything obvious.  I was feeling guilty though that I realise I didn't check with the latest version before opening the Issue (I had 0.14.1 installed) - but just to confirm I have now updated and can reproduce with 0.15.0 as well.


I guess it's something you need to be quite unlucky to run into, so I'm sure not top priority :).  I have worked around it the obvious way (just bumped my `serial` crate to 0.3.2).

Thanks for a super useful tool, by the way (even if I do swear every time it makes my CI/CD pipeline fail ;-).)

crepererum:
Is this technically the same issue as #232? So the common issue is: "non-crates.io crates should be excluded even if the same name exists in crates.io" where "non-crates.io crates" are:

- part of the workspace (#232)
- in a private registry (this issue here)
- (probably also) crates specified via GIT repo or relative file path

tarcieri:
Aha yes, that is definitely a similar issue.

[...]

PR #502, opened 2022-01-07
Fix clippy warnings

tmfink:
Removed `Query` "package_scope" field since it is unused and TODO
comment related to the field.


tarcieri:
I'd rather just tell clippy to ignore it.

This is removing work needed to complete issues like #408

tmfink:
> I'd rather just tell clippy to ignore it.
> 
> This is removing work needed to complete issues like #408

Sure--I can prefix a '_' to get clippy/cargo to suppress the warning. Would you like to leave the "TODO" comment?

PR #516, opened 2022-03-05
Auto generate platforms crate

### LINKED ISSUE #513: Auto-generate the `platforms` crate
`rustc --print=target-list` and `rustc --print=cfg --target=...` now provide all the information encoded in the `platforms` crate and more, except for support tiers.

Support tiers appear to be defined in the documentation, in Markdown tables. This should be feasible to parse as well:
https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support.md

We could use these two data sources to auto-generate the contents of the `platforms` crate, and never have to worry about the list getting stale anymore.

Shnatsel:
- Implement an auto-generator for `platforms` crate data
 - Generate a new version of it, `3.0.0-alpha.1` using the generator
 - Add two new fields, `target_endian` and `target_pointer_width`
 - Mark the `Platform` struct `#[non_exhaustive]`
 - Update `platforms` crate tests not to depend on data that evolves over time
 - Add a script to re-generate the platforms crate (excluding the README.md)
 - Update the in-tree `rustsec` crate to use `platforms` 3.0
 - Remove hacky `cfg` tricks to report some of the properties of the current platform; superseded by [`current_platform`](https://github.com/Shnatsel/current_platform) crate

Closes #513


Shnatsel:
The remaining CI failures seem to be due to `#[deny(warnings)]` on `rustsec`. How do you want me to deal with that @tarcieri ?

Shnatsel:
Dropping the `cfg` stuff was the right call - I have figured out a way to encode the target triple at compile time, so I'll probably just release that as a standalone crate.

tarcieri:
Whoops, screwed up a rebase. Let's see if I can fix it.

tarcieri:
@Shnatsel can you check if I rebased correctly now? (or rebase and force push yourself)

tarcieri:
@Shnatsel okay, we're to an error I think you mentioned before. This advisory:

https://github.com/rustsec/advisory-db/blob/main/crates/time/RUSTSEC-2020-0071.md

...contains a list of bogus OSes. It doesn't like capital-R redox, and there's "rolaris" where ostensibly it should be "solaris".

I think we should just remove that OS list from that advisory? That's not how it's intended to be used.

Shnatsel:
Yeah, there's a bunch of bogus platforms there. We should change the advisory to conform to the actual list of platforms we support.

Shnatsel:
I've fixed the problematic advisory in https://github.com/rustsec/advisory-db/pull/1222

Shnatsel:
CI is now green.

Issue #466, opened 2021-11-05
Problems to install `cargo-audit`, it throws seg-fault


carlosb1:
I was a similar error, for this, but I can not find it.

For a docker image `rust:alpine`, I got this if I install `audit`
```
/albrs-antijamming_rs-lib # cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
Segmentation fault (core dumped)
```
I apply these commands in the docker image:

```
        - apk update && apk upgrade && apk add --no-cache bash git openssh libc-dev openssl-dev
        - rustup component add rustfmt
        - rustup component add clippy
        - cargo install cargo-audit
```
then If I run `cargo audit, it throws the error`

For running the docker image, I apply this:
```
docker run -it --volume=/home/myname/myproject/mylib-lib:/mylib-lib --workdir="/mylib-lib" --memory=4g --memory-swap=4g --memory-swappiness=0 --cpus=4 --entrypoint=/bin/sh rust:alpine
WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
```
Some idea?

tarcieri:
I'm guessing this is something with libgit2.

The best way to help debug this would be to launch `cargo audit` under `gdb` or `lldb` and obtaining a backtrace (e.g. `bt`) so we can see where it's occurring

valkum:
This is my backtrace from using cargo-deny which uses rustsec's Repository::fetch
Installed: 
libgit2: 1:1.2.0-1 (from arch) which should be libgit2.so version 1.2-64
openssl 1.1.1.l-1 

```
#0  0x00007ffff7ad62c3 in SSL_get_peer_certificate () from /usr/lib/libssl.so.1.1
#1  0x00007ffff7f42db5 in ?? () from /usr/lib/libgit2.so.1.2
#2  0x00007ffff7f5034f in ?? () from /usr/lib/libgit2.so.1.2
#3  0x00007ffff7f51eab in ?? () from /usr/lib/libgit2.so.1.2
#4  0x00007ffff7f4fe3c in ?? () from /usr/lib/libgit2.so.1.2
#5  0x00007ffff7f542d6 in ?? () from /usr/lib/libgit2.so.1.2
#6  0x00007ffff7f56739 in ?? () from /usr/lib/libgit2.so.1.2
#7  0x00007ffff7f54a8c in ?? () from /usr/lib/libgit2.so.1.2
[...]

PR #522, opened 2022-04-14
Bump git2 dependency to 0.14

ahollmann:
All other changes were necessary to enable the version bump of git2.

This should fix: https://github.com/rustsec/rustsec/issues/466
This will unblock: https://github.com/EmbarkStudios/cargo-deny/issues/410


tarcieri:
Merged as [cf366ea](https://github.com/rustsec/rustsec/commit/cf366ea3755bb4c31a254d18618870abfb6f729d)

PR #525, opened 2022-04-16
tree-wide: update all the dependencies

liushuyu:
This is a relatively large pull request that upgrades all the dependencies to their latest version. Due to breaking changes in the `abscissa_core` crate, a lot of framework-related changes are made.

The changes could be summarized as follows:

### Changes related to `abscissa_core`

- Use `clap` instead of `gumdrop`. `abscissa_core` no longer bundles a command-line parser, and the upstream project recommends `clap` 3.x for this task.
- As a result of the last change, the command-line parser has been rewritten. To make `clap` happy, a new parsing error was added in `platforms/src/error.rs`
- The base framework has been adapted to the new structure in `abscissa_core`.

### Changes related to `askama`

- Fixed some reference passing issues in `admin/src/web.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,24 +4,22 @@ version = 3
 
 [[package]]
 name = "abscissa_core"
-version = "0.5.2"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


tarcieri:
Thank you for doing this!

There are a few other things I'd like to fix up but I can do that in a followup PR.

PR #528, opened 2022-04-23
refresh `platforms` crate

Shnatsel:
I've run `regenerate-platforms-crate.sh` and committed the result now that a new Rust release is out


tarcieri:
Would be good to run this after #527 is merged

Shnatsel:
I'll wait until #527 is merged then.

I literally only ran `rustup update nightly` and then `./regenerate-platforms-crate.sh`, so it's easier to just re-run that than to resolve conflicts.

PR #531, opened 2022-04-23
Semver compliant platforms

Shnatsel:
- Make individual platform entries `pub(crate)` (a list is still accessible through `ALL`, but platforms can be removed from it)
- Add comments on data sources and stability
- Re-export {Arch, Endian, Env, PointerWidth, OS} for prettier docs

### DIFF
--- a/cargo-audit/src/commands/audit.rs
+++ b/cargo-audit/src/commands/audit.rs
@@ -11,7 +11,7 @@ use crate::{
 use abscissa_core::{config::Override, terminal::ColorChoice, FrameworkError};
 use clap::Parser;
 use rustsec::database::scope;
-use rustsec::platforms::{Arch, OS};
+use rustsec::platforms::target::{Arch, OS};
 use std::{path::PathBuf, process::exit};
 
 #[cfg(feature = "fix")]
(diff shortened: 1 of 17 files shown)


tarcieri (approved):

Shnatsel:
I'm not sure if re-exporting `{Arch, Endian, Env, PointerWidth, OS}` as-is is a particularly good idea. They end up being exported twice, which is probably not ideal?

But without them on the docs.rs frontpage the docs look kinda barebones.

Shnatsel:
Should I perhaps un-`pub` the `target` module?

tarcieri:
Sounds good

Shnatsel:
This does require changes in clients, but we're bumping semver anyway :shrug: 

Does this look good or should I bring `target` back?

Shnatsel:
I decided to bring `target` back and just re-export for the enums to show up on the docs.rs frontpage.

This also simplifies migration for 2.0 users.

tarcieri:
@Shnatsel any other changes you'd like, or should I cut a release?

Shnatsel:
`platforms` 3.0 looks good to go, I'm not aware of any other issues with it.

PR #533, opened 2022-04-23
Cleanup `Cargo.lock` files across workspace

tarcieri:
Removes stale `Cargo.lock` files leftover from when before all of the crates in this workspace were merged into a single repo.

Also moves `platforms/markdown-table-gen` out of the toplevel workspace, and ignores lockfiles for both it and `platforms/platforms-data-gen`.


06kellyjac:
> ..., it is recommended that all binaries check in their Cargo.lock.

`cargo-audit` is a binary so it should really have kept it's `Cargo.lock` file IMO

cc: @tarcieri

tarcieri:
The intent of this PR was to remove the leftover crate-specific Cargo.lock files after combining the repo, now that all of the crates are in a workspace.

However, I see it also removed the toplevel lockfile, which wasn't intended.

PR #541, opened 2022-04-24
rustsec: add `[advisory.source]`; remove old DB scopes

### LINKED ISSUE #408: `cargo audit` incorrectly checks packages in non-crates.io registries

tarcieri:
Fixes #408

Adds a new optional `source` field to the advisory metadata: a URL which identifies the source of the package, e.g. the registry where a vulnerable package is published. These map to the `source` field of Cargo.lock.

The default is crates.io:

    registry+https://github.com/rust-lang/crates.io-index

The `db::Query` type now considers `Package::source`, checking if it's a registry and if so, checking if the advisory's `source` URL matches the `source` of the package. This fixes #408: it ignores packages in Cargo.lock which don't come from a registry (e.g. git, local), and if they do come from a registry, makes sure it's the expected one.

It also enables filing advisories for 3rd party crate registries, which would mainly be useful if someone spun up an internal RustSec DB, or potentially in a future where there are noteworthy public 3rd party registries in use other than crates.io.

This commit also removes the previous `db::scope::{Registry, Package}` originally added in 6f56bfc/0f0a777 for a few reasons:

- The new implementation "just works" without any additional configuration on the part of the user and accomplishes the same functionality. Convention over configuration!
- The legacy `db::scope` system, despite its complexity, was mostly dead code and not actually ever considered in the `Query` (in fact `clippy` just recently started failing due to that). The only functionality it actually provided was allowing a user to configure that they only want to consider packages in their Cargo.lock whose source is crates.io
- In addition to that, the `db::scope` system duplicated functionality already available in the `SourceId` type, resulting in needless complexity


rustsec/src/advisory/metadata.rs
```diff
@@ -63,13 +63,17 @@ pub struct Metadata {
...
     /// Additional reference URLs with more information related to this advisory
     #[serde(default)]
     pub references: Vec<Url>,
 
+    /// URL with an announcement (e.g. blog post, PR, disclosure issue, CVE)
+    pub url: Option<Url>,
+
+    /// URL for the source registry where the vulnerable package is published
+    /// (defaults to crates.io, i.e. `registry+https://github.com/rust-lang/crates.io-index`)
+    pub registry: Option<Url>,
```
tarcieri, line 75: One question: should this be a `Vec<Url>` (or perhaps `Option<Vec<Url>>`?)

I think `Option<Url>` is probably okay for now. Perhaps there will be a future where people are syndicating packages to multiple repositories, but this doesn't seem to be a particularly common thing with any existing packaging system.

Another alternative to consider is a `pub source: Option<SourceId>` which would permit sources other than package registries. I actually tried that originally and it seemed to overcomplicate things, and I couldn't really envision any use cases for filing advisories for anything other than packages published to registries.

rustsec/src/database/query.rs
```diff
@@ -40,9 +43,6 @@ pub struct Query {
 
     /// Query for informational advisories
     informational: Option<bool>,
-
-    /// Scope of packages which should be considered for audit
-    _package_scope: Option<scope::Package>,
```
tarcieri, line 45: Dead code

rustsec/src/database/query.rs
```diff
@@ -70,19 +70,16 @@ impl Query {
...
-    ) -> Self {
-        self.package = Some(package.into());
-        self.version = Some(version.into());
+    /// Provide a package and use all of its attributes as part of the query
+    pub fn package(mut self, package: &Package) -> Self {
+        self.package_name = Some(package.name.clone());
+        self.package_version = Some(package.version.clone());
+        self.package_source = package.source.clone();
         self
     }
```
tarcieri, line 84: Should note these are renamed in the changelog

rustsec/tests/database.rs
```diff
@@ -27,31 +22,13 @@ fn vulnerabilities_default() {
...
-        Lockfile::load(lockfile_path).expect("Should find the lock file in support folder.");
-    let db = DEFAULT_DATABASE.lock().unwrap();
-
-    let vuln_public =
-        db.query_vulnerabilities(&lockfile, &Query::crate_scope(), scope::Registry::Public);
-    assert_eq!(vuln_public.len(), 0);
-
-    let vuln_all = db.query_vulnerabilities(&lockfile, &Query::crate_scope(), scope::Registry::All);
-    assert_eq!(vuln_all.len(), 1);
-}
```
tarcieri, line 57: This is the only functionality provided by the previous `db::scope` system.

I'll add replacement test(s) to exercise the registry scoping rules, testing that an advisory doesn't match if it's for a `registry` different from `source`.

Shnatsel:
How does this interact with vendored dependencies? Do we just ignore local and `git` crates by default?

I suspect that that vendored dependencies are more common than private registries.

tarcieri:
Per #408, previously we had potential false positives in these situations if the name of a local crate clashed with crates.io.

We have no way of knowing in these cases whether or not a crate is a vendored version of a crates.io crate or not, or if the versioning scheme used matches.

Shnatsel:
Ah, yeah. So the question is, do we prefer false positives or false negatives?

For example, if I clone a git repo and run `cargo audit` on it, and there was an advisory for *this very project,* it will not be reported if we ignore local crates by default. I'm not sure that's a good default behavior.

tarcieri:
The corollary problem is if we do treat a local package as if it were a package on crates.io, if someone runs `cargo audit` on a package published to a 3rd party repo which happens to have the same name as the crates.io package, they'll get a false positive.

But #408 is more than a "false positive" problem: the core problem is we currently have no solution for 3rd party package registries, and the previous matching logic is naive and didn't consider the package's source at all.

The RustSec Advisory DB is specifically advertised as "a repository of security advisories filed against Rust crates published via https://crates.io/". This PR makes it actually does what it says on the label, while also making it possible for there to be 3rd party advisory DBs published against 3rd party repos.

Shnatsel:
>  Regarding `cargo audit` not matching a local crate under this behavior, the explanation is straightforward: the local git copy of the crate isn't published to crates.io.

But there is no way to run `cargo audit` on the copy published to crates.io, is there? So there's now no way to run an analysis on the entire dependency tree.

Could we perhaps treat everything *local* as coming from crates.io by default, and let people opt out of that behavior instead?

tarcieri:
I'm not sure what behavior you actually want for local repositories.

Let's say someone has a local git clone of a crates.io package which has a known vulnerability. They create a fix for the vulnerability, commit it, and open a PR. CI runs, but it happens to run `cargo-audit` as well. Even though the PR is to fix the vulnerability, if local packages are treated as crates.io packages, the audit fails, because the crates.io copy is vulnerable.

To me it would be surprising to consider local packages as if they're crates.io packages. The behavior in this PR uses the same sort of resolution Cargo does i.e. a particular package release is a 3-tuple of `name, version, source`, which is what I would respect.

Shnatsel:
Since we disagree on what the least surprising behavior is, perhaps we should run a community poll? Maybe on RustSec twitter or some such?

tarcieri:
Can you give an example of where you'd actually find that behavior helpful?

I've given two examples of where it's undesirable.

rustsec/src/advisory/metadata.rs
Shnatsel: GHSA allows publishing advisories about Git repos, without requiring a package - but they don't include those in their curated set AFAIK. But if we support that, we also need a way to specify affected versions as git commit ranges - which OSV does, but I think will vastly overcomplicate things for us.

rustsec/tests/database.rs
Shnatsel: SGTM

rustsec/src/advisory/metadata.rs
tarcieri: There isn't currently, although it's easy enough to do today by running `cargo audit` multiple times with different `-d` flags

tarcieri: Re: git, I could see the use case there (e.g. a company has an internal monorepo of crates sourced from git rather than a registry, and wants to have an internal advisory DB for it, possibly in the same repo)

tarcieri:
A note on the above: vendored crates are a bit harder, because `source` is unset in Cargo.lock for crates in the current workspace.

So we'd need something like `--match-local` and/or a corresponding `audit.toml` option to match crates whose `source` is unset

tarcieri:
I've added an `--ignore-source` CLI arg and `ignore-source` key in audit.toml which can be used to enable the previous behavior.

I strongly believe that honoring the source should be the default behavior. A name alone does not uniquely identify a package, and that means the previous behavior was naive and false-positive ridden. This is a major impediment for anyone hosting an internal registry and is only going to get worse as more 3rd party registries become available.

Regarding the ability to alias a source for the purpose of handling mirrors, that's a bit trickier and something I'd rather add as a follow-up to this.

tarcieri:
Also note: I plan on still leaving this open for at least another week. Feel free to bikeshed the `--ignore-source` name or suggest alternative ways it could be structured.

I'd like to experiment (again) with replacing `advisory.registry` with `advisory.source` (or thereabouts). We can have the lint enforce it's a registry for now, but it's much more futureproof if we'd like to eventually add support for things like filing advisories against git-sourced dependencies.

Shnatsel:
If we're changing the default behavior to ignore local packages and packages from third-party registries, perhaps we should at least print a warning that those packages were not audited?

FWIW this is what `cargo supply-chain` does, to make it clear what is and isn't reported.

Something like:
```
The following packages were not audited because they do not come from crates.io: 
 A, B, C
Use the --ignore-source flag to include them in the audit.
```

tarcieri:
That's a good idea, but I think it only makes sense when `source` is set to `Some(...)`, e.g. 3rd party registries and git dependencies.

Otherwise it will always alert for every crate in the local workspace, and at present we have no way to suppress warnings. This would not only be needlessly spammy, but it would make it impossible to use `--deny warnings` as there would always be at least one warning.

Shnatsel:
Yeah, I think it shouldn't be covered by `--deny warnings`

 I definitely want the list of not-audited local crates to be printed in situations when people use vendored crates. The default of not auditing local crates would lead to false negatives, and that is not immediately apparent from the tool, leading to insidious false negatives.

tarcieri:
To support that, we'll need to add some sort of feature for that which can map certain local paths containing vendored crates to the registry the crates were vendored from.

That's something I'd prefer to keep out-of-scope from this particular PR.

Shnatsel:
Yes, so for now I want to just inform the end user that these crates are not being audited, and have them pass `--ignore-source`, because there is currently no better solution for that use case.

tarcieri:
I can see if `source` is set to `Some(...)` in the event that a dependency is referred to by `path` but not a member of the local workspace.

If so, they'll "just work" with source-based alerting for unaudited crates.

rustsec/src/advisory/metadata.rs
tarcieri: I've changed the field to `source` and made it store a full `SourceId`. This makes it possible to potentially support git-based sources in the future.

For now I suggest linting the field to ensure it's `registry+https://`

tarcieri:
The PR has been updated to use a `source` field which stores a full `SourceId`. This makes it potentially possible to file advisories against any source type.

For now I think it makes sense to use a lint to constrain it to package registries.

PR #551, opened 2022-05-02
CI: use `rust-cache` for caching

tarcieri:
Uses our fork of the upstream project:

https://github.com/rustsec/rust-cache


.github/workflows/cargo-audit.yml
```diff
@@ -40,24 +27,17 @@ jobs:
           - macos-latest
           #- windows-latest # TODO(tarcieri): debug Windows build timeouts
         toolchain:
-          - 1.57.0 # MSRV
+          #- 1.57.0 # MSRV
```
tarcieri, line 30: For some reason it's not working when an explicit Rust release version is provided.

I opened an upstream issue: https://github.com/Swatinem/rust-cache/issues/57

tarcieri:
Closed in favor of #929

PR #565, opened 2022-05-17
Remove duplicate Keywords entry in advisory page

amousset:


admin/src/web/templates/advisory-content.html
```diff
@@ -214,15 +214,6 @@ <h1>
       </dd>
       {% endif %}
 
-      {% if advisory.metadata.keywords.len() > 0 %}
```
amousset, line 217: duplicate of https://github.com/rustsec/rustsec/pull/565/files#diff-93e34cc83a4684122d90fda4b37479c3886c2f3c24108713d7086a88c8c31363R74

PR #596, opened 2022-06-29
Encode version into V3 lockfiles

abrisco:
V3 lockfiles currently do not include their version when encoded causing them to be mistaken for V2 lockfiles in some cases. This PR fixes that by lockfiles V3 or newer encode `version = #`.


tarcieri:
Good catch, thanks!

PR #599, opened 2022-06-30
Add missing newline for encoded version variables.

abrisco:
Currently lockfiles are encoded as the following due to a missing newline in https://github.com/rustsec/rustsec/pull/596:
```toml
# This file is automatically @generated by Cargo.
# It is not intended for manual editing.
version = 3
[[package]]
```

This change fixes the encoding to correctly match Cargo generated `Cargo.lock` files
```toml
# This file is automatically @generated by Cargo.
# It is not intended for manual editing.
version = 3

[[package]]
```


tarcieri:
Thanks! Perhaps it would be good to add a test that it round-trips a lockfile with a `version` back to the original encoding?

cargo-lock/tests/examples/Cargo.lock.v1
```diff
@@ -1,6 +1,5 @@
-# Example non-trivial `Cargo.lock` file.
```
abrisco, line 1: I'm not sure about the V1 lockfiles here. Did they normally get generated with this unique header?

abrisco (commented):
> Thanks! Perhaps it would be good to add a test that it round-trips a lockfile with a version back to the original encoding?

How's this?

tarcieri (approved):

cargo-lock/tests/examples/Cargo.lock.v1
abrisco: Thanks! that's exactly what I did, just wasn't sure if it was the correct thing to do 😄

PR #620, opened 2022-07-31
Fix 619 with rustsec-admin lint --skip-namecheck

### LINKED ISSUE #619: Admin: Handling of blacklisted crates
rustdecimal was and is a malicious crate that was blackholed by crates.io

https://github.com/rustsec/advisory-db/pull/1234

This means our linter in CI fails due this not existing in crates.io

https://github.com/rustsec/advisory-db/runs/6375132406?check_suite_focus=true#step:5:74

Ideally need to fix it in a way that doesn't affect lints normally but can warn without error under this circumstance

pinkforest:
Fixes #619

EDIT: Old design proposal (crates_io metakey) has been removed from the Issue and edited in this PR for the final version

### DIFF
--- a/admin/src/commands/lint.rs
+++ b/admin/src/commands/lint.rs
@@ -17,6 +17,9 @@ pub struct LintCmd {
         help = "filesystem path to the RustSec advisory DB git repo"
     )]
     path: Vec<PathBuf>,
+
+    #[clap(long, help = "Skip name check comma separated crates list")]
+    skip_namecheck: Option<String>,
 }
 
 impl Runnable for LintCmd {
@@ -27,7 +30,7 @@ impl Runnable for LintCmd {
             _ => unreachable!(),
         };
 
-        let linter = Linter::new(&repo_path).unwrap_or_else(|e| {
+        let linter = Linter::new(&repo_path, self.skip_namecheck.to_owned()).unwrap_or_else(|e| {
             status_err!(
                 "error loading advisory DB repo from {}: {}",
                 repo_path.display(),
--- a/admin/src/linter.rs
+++ b/admin/src/linter.rs
@@ -118,16 +118,9 @@ impl Linter {
         Ok(())
     }
 
-    /// Skip crates.io name check if the name has been blackholed issue/619
-    fn skip_name_exists_on_crates_io(&self, advisory: &rustsec::Advisory) -> bool {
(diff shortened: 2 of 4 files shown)


tarcieri:
The advisory format is not particularly coupled to crates.io and designed with the eventual goal of supporting other registries, so if we do add a key like this, it should be generic and usable with any registry.

As an alternative to trying to encode this information directly into advisories, which requires modifying the `rustsec` crate with changes to the public API, I think we can just add a CI option to skip lint for certain advisories, which can be implemented entirely within `rustsec-admin`.

pinkforest:
Ah good point

pinkforest:
Oh wait that doesn't make sense.

As when we merge that advisory the CI still continues to run with error when the advisory is stored

tarcieri:
A CLI option seems fine to me for now. We can merge config changes first, then the advisory.

There's only one crate this impacts for now, so I don't think managing a large list is a pressing issue. And keeping configuration entirely on the `rust-admin`/CI side means we can change it at will down the road without those changes impacting the public API of the `rustsec` crate.

pinkforest:
So it's now this:

`rustsec-admin lint databasepath --skip-namecheck rustdecimal;anothercrate`

--help produces:
```
$ date;rustsec-admin lint ../../ --help
Sun 31 Jul 2022 23:28:47 AEST
rustsec-admin-lint 
lint Advisory DB and ensure is well-formed

USAGE:
    rustsec-admin lint [OPTIONS] [PATH]...

ARGS:
    <PATH>...    filesystem path to the RustSec advisory DB git repo

OPTIONS:
    -h, --help                               Print help information
        --skip-namecheck <SKIP_NAMECHECK>    Skip name check comma separated crates list
```

admin/src/commands/lint.rs
```diff
@@ -17,6 +17,9 @@ pub struct LintCmd {
         help = "filesystem path to the RustSec advisory DB git repo"
     )]
     path: Vec<PathBuf>,
+
+    #[clap(long, help = "Skip name check comma separated crates list")]
+    skip_namecheck: Option<String>,
```
tarcieri, line 22: `clap`/`structopt` at least used to have first-class support for comma-separated arguments, although I'm not sure what the status is with `clap` v3 (the `clap_derive` docs are 404ing, unfortunately)

PR #631, opened 2022-08-04
Deprecate yanked

pinkforest:
**When I was moving http-tower to 2022**
https://github.com/rustsec/advisory-db/pull/1319
https://github.com/rustsec/advisory-db/pull/1320

Also happened here:
https://github.com/rustsec/advisory-db/pull/1310

Linter still requires **yanked = true**
`malformed content in [metadata]: Advisories with the withdrawn field must also set yanked = true`

## Deprecation of `yanked`

This will only error in case where `withdrawn` has not been set when `yanked` was still set, allowing deprecation of `yanked` field

We can then remove `yanked` from all the existing advisories and then we can further take the field out from linter completely after

### DIFF
--- a/rustsec/CHANGELOG.md
+++ b/rustsec/CHANGELOG.md
@@ -4,10 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.27.0 (TBD)
+## 0.26.1 (TBD)
 ### Added
 ### Changed
-- Deprecate `yanked` (#TBD])
+- Deprecate `yanked` ([#631])
+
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

rustsec/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.27.0 (TBD)
```
tarcieri, line 7: Can probably just call this...

```suggestion
## 0.26.1 (TBD)
```

It's a behavior change and not one that concerns end users.

pinkforest: Yeah suspected, took the opportunity to fix the link as well :)

PR #635, opened 2022-08-10
Binary scanning MVP

Shnatsel:
This is an MVP, to have some support for `cargo auditable` on launch.

Eventually I'd like to extend this with `--dir` parameter to scan an entire directory, because scanning files one by one causes us to reload the DB every time, which is slow. And once that's in place, make `cargo audit bin` scan the programs installed by Cargo instead of requiring `--file` or `--dir` parameters.

I'm very open to constructive criticism here.


cargo-audit/src/auditor.rs
```diff
@@ -147,6 +147,21 @@ impl Auditor {
 
         self.presenter.before_report(lockfile_path, &lockfile);
 
+        self.audit(&lockfile)
+    }
+
+    #[cfg(feature = "binary-scanning")]
+    /// Perform an audit of a binary file with dependency data embedded by `cargo audit`
```
amousset, line 154:
```suggestion
    /// Perform an audit of a binary file with dependency data embedded by `cargo auditable`
```

Shnatsel:
Also, should `--bin` be a flag or a subcommand? We have to decide now because it would be difficult to change later.

I have used it as a flag because I wanted to reuse most flags from the regular invocation, but I am also slightly concerned about confusion with Cargo's `--bin` parameter which behaves differently.

amousset:
> And once that's in place, make cargo audit --bin scan the programs installed by Cargo instead of requiring --file or --dir parameters.

Do you mean `$HOME/.cargo/bin` or binaries in the `target/` dir of current project?

I feel like a common use case for such a command is to audit production artifacts (binaries in containers, etc.) which don't embed dev tooling, so it seems to me that the `--bin` case is less linked to a Rust dev/build env and more to arbitrary file audits. I would expect `cargo audit --bin /usr/local/bin/progam` to work (but that's obviously a very minor nitpick).

Shnatsel:
I meant `$HOME/.cargo/bin`

> I would expect `cargo audit --bin /usr/local/bin/progam` to work

Yeah, that would make sense, but there's already an `-f` flag for specifying the textual file to scan. Should I make `--bin` accept an argument and conflict with `-f`?

Then again, I would intuitively expect `cargo audit --bin /usr/bin/progam /usr/bin/other-progam` to work, and that would require parsing free-standing arguments if `--bin` is passed, instead of getting the value passed to `--bin`.

I think we're not using free-standing arguments for anything, so I could implement that. Is that a good idea?

Shnatsel:
I've reworked the CLI to accept multiple binaries to scan at once. This speeds up scanning multiple binaries dramatically, since the database only needs to be loaded once.

Shnatsel:
I don't understand what's up with CI. The error is:

```
Run cargo build --no-default-features --release --target wasm32-unknown-unknown
    Updating crates.io index
error: failed to select a version for the requirement `libgit2-sys = "=0.13.4"`
candidate versions found which didn't match: 0.13.2+1.4.2, 0.13.1+1.4.2, 0.13.0+1.4.1, ...
location searched: crates.io index
required by package `git2 v0.14.4`
    ... which satisfies dependency `git2 = "=0.14.4"` of package `rustsec v0.26.2 (/home/runner/work/rustsec/rustsec/rustsec)`
    ... which satisfies path dependency `rustsec` of package `rustsec-admin v0.8.1 (/home/runner/work/rustsec/rustsec/admin)`
Error: Process completed with exit code 101.
```

But it works fine for me locally on 1.56.0, and there seem to be no uncommitted Cargo.lock files in the tree (checked with `git status --ignored`). Weird.

tarcieri:
It's probably a cached crates.io index (which, for some reason, cargo doesn't choose to update even if it's quite old) which doesn't contain that particular release.

You can try disabling the caching and seeing if that fixes the issue.

Shnatsel:
How do I disable caching? Remove the cache job from the CI specification of this branch?

tarcieri:
Yes. You can at least see if that fixes the problem.

Shnatsel:
Most jobs don't seem to use cache as it is. I have dropped it from the one job that did have it, but that had no effect.

Shnatsel:
Dropping Cargo.lock from the tree did help, but I don't understand why.

tarcieri:
Perhaps try `--locked`?

Shnatsel:
To which command should I pass `--locked`?

tarcieri:
`cargo`: https://doc.rust-lang.org/cargo/commands/cargo.html?highlight=locked#manifest-options

See also: `--frozen`

Shnatsel:
Well, writing tests has caught a genuine bug: the --db parameter is not honored by the `bin` subcommand. That's a blocker.

PR #641, opened 2022-08-15
Add search field for IDs and packages

### LINKED ISSUE #637: Add a search field on rustsec.org
As explained in https://github.com/rustsec/advisory-db/issues/1253 it is currently difficult to find an advisory marthing a specific CVE id (or some other details).

A way to address this in a general manner would be to add a search feature on rustsec.org. There are two main options:

* The easiest one is to use an third-party service (like [Algolia Docsearch](https://docsearch.algolia.com/) which is free for open-source projects), which can provide indexing and search results of good quality.
* The other option is to build a static index along with the site and to use a library like [lunr.js](https://lunrjs.com) to make the search.

amousset:
Add a very simple search feature, only making exact searches over package names and IDs. I think it would already cover a lot of needs. What it does:

* The user enters a search query and submits the search
* If the term is a known package, redirect to the already existing index of vulnerabilities for this package (like https://rustsec.org/packages/openssl-src.html)
* If the term is a known ID, and this ID appears in only one advisory, redirect to this advisory directly
* Else, go to the new search results page:
  * If the term is a known ID with several linked advisories (i.e. id, alias or related), list the linked advisories
  * Else display a "no results" message

image
image

The index is built with the site, only contains a list of packages and a map of `ID -> linked IDs` (and weights 36k uncompressed).

Fixes #637.

_Side note: The js and css in this would probably make a front-end developer cry._


amousset:
@tarcieri Do you think it is functionally adequate or would we need a fancier result page or richer search features?

Shnatsel:
I just have one request: use "look up" instead of "search". This assumes exact matches, but I'm used to search matching substrings.

I'm happy to merge this once this one UI change is done. I'm particularly happy to see the CVE and GHSA ID lookup working - thanks a lot for that!

PR #645, opened 2022-08-15
Revert "build(deps): bump comrak from 0.13.0 to 0.14.0"

Shnatsel:
Revert rustsec/rustsec#608 to see if that fixes tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -195,21 +195,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bit-set"
-version = "0.5.2"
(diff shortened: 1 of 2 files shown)


tarcieri:
What did this break?

Shnatsel:
https://github.com/rustsec/rustsec/runs/7838276597?check_suite_focus=true

tarcieri:
Huh, weird. Wonder why it passed the first time.

I think the error is coming from `serde_json`? https://github.com/serde-rs/json/blob/dab5ed3ee97cef5e2b796513f8d9e4c7416e44bf/src/error.rs#L293

PR #655, opened 2022-08-20
Update comrak to 0.14

amousset:
Re-try update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1709,7 +1709,7 @@ dependencies = [
 
 [[package]]
 name = "rustsec"
-version = "0.26.1"
+version = "0.26.2"
 dependencies = [
  "cargo-edit",
  "cargo-lock",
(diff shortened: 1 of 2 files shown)


tarcieri:
Looks like it failed again:

https://github.com/rustsec/rustsec/runs/8025899590?check_suite_focus=true#step:7:117

I'm a bit confused in that it's only failing on macOS, and the error seems to be JSON-related:

```
---- advisories_found_but_ignored_json stdout ----
thread 'advisories_found_but_ignored_json' panicked at 'called `Result::unwrap()` on an `Err` value: Error("EOF while parsing a value", line: 1, column: 0)', cargo-audit/tests/acceptance.rs:70:35
```

...which seems to come from here:

https://github.com/serde-rs/json/blob/dab5ed3ee97cef5e2b796513f8d9e4c7416e44bf/src/error.rs#L293

tarcieri:
Even more annoying: I'm unable to reproduce this locally using Rust 1.57.0 on macOS.

tarcieri:
Opened a tracking issue: #660

We can revert this again, although I'd really like to figure out what's causing the problem.

Issue #644, opened 2022-08-15
Add a command to consolidate database against external sources


amousset:
We could add a command to:

* detect missing aliases/related IDs (based on CVE and/or GHSA data), and maybe open pull requests automatically to add them
* detect advisories present as GHSA but not in advisory-db

amousset:
A simple PoC using osv data https://gist.github.com/amousset/4585cf0d59d1f243536af70081fdd477

amousset:
This produces an output looking like:

```
[...]
Missing alias GHSA-fg7r-2g4j-5cgr in RUSTSEC-2021-0124
Missing alias GHSA-fhvj-7f9p-w788 in RUSTSEC-2020-0034
Missing alias GHSA-wrvc-72w7-xpmj in RUSTSEC-2019-0026
Missing alias GHSA-pphf-f93w-gc84 in RUSTSEC-2020-0111
Missing alias GHSA-wgx2-6432-j3fw in RUSTSEC-2020-0025
Missing alias GHSA-5325-xw5m-phm3 in RUSTSEC-2021-0074
Missing alias CVE-2020-36511 in RUSTSEC-2020-0153
Missing alias CVE-2020-36514 in RUSTSEC-2020-0155
Missing advisory GHSA-pvh2-pj76-4m96 in advisory-db
Missing advisory GHSA-v935-pqmr-g8v9 in advisory-db
```

This could be used to automate aliases update (GHSA and CVE). Advisories present as GHSA- but not RUSTSEC- are not actionable immediately but it could be useful to be aware of them anyway.

@Shnatsel what is your opinion about this?

Shnatsel:
We really need something like this! Thanks for writing it!

I have been [experimenting](https://github.com/rustsec/rustsec/tree/ghsa) with this as well, although I was querying the GHSA API, not OSV data. I was originally blocked by GHSA not including references to RustSec advisories, but it has been fixed since.

I am happy with detection via the regex on "references" field, that's also how I implemented it. I'm afraid that fetching the entire ZIP file might get unwieldy over time, and querying an API since the last update would incur less traffic - but it also requires a lot more complexity, so I'm fine with just fetching a ZIP for now.

We will need both to import the IDs, and to import advisories that we don't presently carry. The GHSA export to OSV is rather poor right now, with a lot of data lost or exported incorrectly (especially version specifications), so I'm afraid we'll have to use the GHSA API for anything more than IDs.

Shnatsel:
[...]

PR #656, opened 2022-08-21
Add a command to synchronize advisory data from osv.dev/GHSA

amousset:
Result in https://github.com/rustsec/advisory-db/pull/1693.

Refs: https://github.com/rustsec/rustsec/issues/644

This PR implements:

* Alias id update from OSV advisories
* Listing OSV advisories we should consider importing into RustSec

The end goal is to be able to automatically open pull requests with the advisories updates and drafts from missing ones. For now I plan to run it regularly manually until we're confident with the result.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2991,6 +2991,7 @@ dependencies = [
  "termcolor",
  "thiserror",
  "toml 0.7.8",
+ "toml_edit 0.19.11",
  "xml-rs",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2991,7 +2991,7 @@ dependencies = [
  "termcolor",
  "thiserror",
  "toml 0.7.8",
- "toml_edit 0.19.11",
+ "toml_edit 0.19.15",
  "xml-rs",
 ]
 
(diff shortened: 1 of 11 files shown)


cargo-audit/src/commands/audit.rs
```diff
@@ -1,5 +1,7 @@
 //! The `cargo audit` subcommand
 
+#![allow(unused_qualifications)]
```
amousset, line 3: Rust 1.70.0 produces:

```
warning: unnecessary qualification
  --> cargo-audit/src/commands/audit.rs:38:12
   |
38 |     #[clap(subcommand)]
   |            ^^^^^^^^^^
   |
note: the lint level is defined here
  --> cargo-audit/src/lib.rs:21:56
   |
21 | #![warn(missing_docs, rust_2018_idioms, trivial_casts, unused_qualifications)]
   |                                                        ^^^^^^^^^^^^^^^^^^^^^
```

I did not investigate it further for now.

tarcieri: I really need to upgrade `clap` (and `abscissa`)

Shnatsel:
@amousset any reason not to merge this and start running it on CI?

It would be great to have this up and running!

amousset:
It's been running correctly since a long time, I think it makes sense to start automating this indeed!

PR #662, opened 2022-08-27
Relax git2 requirement

kornelski:
Allows using git2 0.15 too


alex:
is there a reason not to just upgrade to 0.15?

tarcieri:
If it compiles with both, I think this requirement is OK.

However it would probably make sense to check it continues to compile with `-Z minimal-versions` if we do this.

kornelski:
git2 depends on libgit2-sys that has a `links` attribute, so only one version can exist in a program. git2 versions are fragmented now, so supporting both helps: https://lib.rs/crates/git2/rev

AFAIK there aren't any real breaking changes between the two versions, so allowing both is harmless.

tarcieri:
Cool, let's try it out

PR #666, opened 2022-08-28
Improve appearance of home page

amousset:
Make the about page more visually appealing as it is also used as homepage. This type of design has the drawback or being less dense but given the level of information displayed I think it makes sense.

Current page:
image

Proposition:
image

<details>
<summary>previous drafts</summary>

</details>


tarcieri:
Looks great!

amousset:
Second iteration
image

amousset:
Third iteration, ready for feedbacks
image

pinkforest:
Looks good! 

Could link to howto unmaintained and contributing.md etc. from there to how to get started with contributing to advisories, but looks good! :octopus: 

I mean those .md files in our advisory repo could be somehow presented in the website perhaps :thinking:

amousset:
> I would perhaps not mention how many advisories there are and just say "browse"

I have no strong opinion about this. I agree this is not really useful, but one small advantage could be to indicate there are "many" advisories and it is hence a "real database".

> And reporting vulnerabilities probably at top

I think more people want to consume the database than report a vulnerability, so I left it at the bottom, but IMO that's actually something that should be accessible from every page, so I would add it to the main menu.

pinkforest:
> I have no strong opinion about this. I agree this is not really useful, but one small advantage could be to indicate there are "many" advisories and it is hence a "real database".

nit: Hmm I don't think we need to make that case .. I styilistically prefer simple :)

> I think more people want to consume the database than report a vulnerability, so I left it at the bottom, but IMO that's actually something that should be accessible from every page, so I would add it to the main menu.

nit: Yeah I thought it was a bit distracted and I thought where is the report button and it took quite bit to find it :)

amousset:
I updated the screenshot

* Address feedback from @pinkforest 
* Added a new "Report Vulnerabilities" page that renders the CONTRIBUTING.md file from advisory-db (this could be done in another PR but I wanted to test the consistency of both changes)

pinkforest:
looks really nice :) :partying_face:

pinkforest:
just one more nit if I may: "RustSec Vulnerability Database" - I think the heading is too long :) cheers

amousset:
> just one more nit if I may: "RustSec Database" could simply work as a heading - I think the heading is too long and may not signal true extent of what we provide e.g. unmaintained and unsound etc. which are not strict / concrete vulns :) Cheers

In this case we would need another sentence below to explain what it means.

pinkforest:
> In this case we would need another sentence below to explain what it means.

nit: yeah we already have that at the bottom.. it's like google search we have the logo at top and explanation at bottom for new users :)

amousset:
I really think we need a one sentence explanation of what it is on the first screen of rustsec.org, RustSec is a bit more niche than Google :smile:

tarcieri:
I like it, although I agree a longer-form description would also be nice

pinkforest:
The long form description is at the bottom ?

Shnatsel (approved):
Looks great! I think the Github tile should be expanded a bit - mention that they our data is available via Dependabot alerts and the API. But that shouldn't block merging this.

amousset:
image
Updated :heavy_check_mark:

PR #675, opened 2022-09-01
Add an icon after external links in advisory content

amousset:
Helps making the difference between internal and external links in the advisories, it is not always obvious.

image

### DIFF
--- a/admin/src/web/static/css/basic.css
+++ b/admin/src/web/static/css/basic.css
@@ -52,6 +52,17 @@ header a, footer a {
   font-weight: 500;
 }
 
+/* Add an icon after external links in advisory content */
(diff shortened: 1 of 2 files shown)


amousset:
Note: The icon comes from [fontawesome](https://raw.githubusercontent.com/FortAwesome/Font-Awesome/6.x/svgs/solid/arrow-up-right-from-square.svg), under CC-BY 4.0, as stated in the comment in the SVG. The linked [licensing docs](https://fontawesome.com/license/free) says:

> Attribution is required by MIT, SIL OFL, and CC BY licenses. Downloaded Font Awesome Free files already contain embedded comments with sufficient attribution, so you shouldn't need to do anything additional when using these files normally.

tarcieri (approved):

PR #679, opened 2022-09-02
Set chrono features and bump chrono to 0.4.22

pinkforest:
So I got rid of one of the references to time 0.1 earlier by getting rid of clap2

But chrono still had default-features which meant it brought "oldtime"

So I got rid of that:

I specified the only feature `Clock` that we need from chrono and gets rid of the `oldtime` that brought time 0.1

For some reason chrono had `serde` feature specified but I don't see what is using it ?

The only usage of chrono is in admin/src/web.rs:
```
admin/src/web.rs:use chrono::{Date, Duration, NaiveDate, Utc};
admin/src/web.rs:    use chrono::NaiveDate;
```

I also bumped .lock to chrono 0.4.22 with update -p libc bump:
```
$ cargo update -p chrono --precise 0.4.22
    Updating crates.io index
      Adding android_system_properties v0.1.5
    Updating chrono v0.4.20 -> v0.4.22
      Adding iana-time-zone v0.1.46
    Updating js-sys v0.3.57 -> v0.3.59
    Updating wasm-bindgen v0.2.80 -> v0.2.82
    Updating wasm-bindgen-backend v0.2.80 -> v0.2.82
    Updating wasm-bindgen-macro v0.2.80 -> v0.2.82
    Updating wasm-bindgen-macro-support v0.2.80 -> v0.2.82
    Updating wasm-bindgen-shared v0.2.80 -> v0.2.82
```

It all builds and tests okay but not sure why the serde feature was defined that seems it wasn't required :thinking:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,6 +66,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "ansi_term"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


amousset (approved):

tarcieri:
@pinkforest the `serde` feature was possibly used at one point, but seems vestigial now

PR #682, opened 2022-09-04
Add license and attribution fields to advisories

amousset:


rustsec/src/advisory/metadata.rs
```diff
@@ -81,4 +82,12 @@ pub struct Metadata {
...
     #[serde(default)]
     pub withdrawn: Option<Date>,
+
+    /// License of the advisory content
+    #[serde(default)]
+    pub license: License,
+
+    /// Attribution for the advisory, required for some licenses
+    #[serde(default)]
+    pub attribution: Vec<String>,
```
amousset, line 92: not sure about the representation here, maybe we want to be able to give a URI too

admin/src/web/templates/advisory-content.html
```diff
@@ -265,4 +265,13 @@ <h1>
 
     <h3 id="description">Description</h3>
     {{ rendered_description|safe }}
+
+    <p id="license" class="license">Advisory available under <a href="https://spdx.org/licenses/{{ advisory.metadata.license.as_str() }}.html">{{ advisory.metadata.license.as_str() }}</a>
```
amousset, line 269: We could restrict this information to advisories not under CC0.

rustsec/src/advisory/license.rs
```diff
@@ -0,0 +1,70 @@
+use crate::Error;
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use std::str::FromStr;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+/// Opaque type representing allowed licenses for advisory content
+pub enum License {
+    CreativeCommonsZero10,
+    CreativeCommonsCc40,
+}
```
tarcieri, line 10: This list isn't extensible.

We could add a catch-all variant like `Other(String)`, and then lint that the license is recognized.

Or just use a `String` and use lints to check it's a valid license.

pinkforest: isn't there SPDX crate that gives these and non-exhaustive form ?

amousset: > This list isn't extensible.

Indeed. If I understand correctly we want to be able to parse all correctly formatted advisories (for example to be able to parse future advisories), and do the validation in the linter as much as possible (to validate db with current tooling). I did not identify this at first, I updated the PR with an `Other` field. 

> isn't there SPDX crate we could use that gives these and non-exhaustive form ?

There is `spdx` (by Embark, used in cargo-deny), and I started using it at first. I ended up considering it did not bring much value as we will need to strictly restrict the allowed licenses in a lint, so it was bringing a dependency for nothing.

amousset: I've hidden the variants to make the underlying type invisible and allow switching to a richer representation later.

rustsec/src/advisory/metadata.rs
amousset: cc attribution doc https://wiki.creativecommons.org/wiki/License_Versions#Detailed_attribution_comparison_chart

amousset: So I think we need to ability to document links.

pinkforest:
Don't think we need a separate license field ?

I mean we could just say source = ["database", ..] .. and then that source triggers both license and attribution automatically ?

I mean the attribution is like the only thing we have to worry right ?

Then linter checks source includes either OSV, GHSA or CVE as valid / supported data sources we / presentation knows how to handle.

Then at presentation layer that generates appropriate attribution in the website and instead of setting license ad-hoc to advisories we can just generally mention "Unlicense or XX where source is YY" where most of the advisories would not have source when we are the primary source.

Additional trouble is recursive attribution but then it can just have multiple sources = [] which just generates multiple attributions.

amousset:
> I mean we could just say source = ["database", ..] .. and then that source triggers both license (if req.) and attribution automatically

Could work and is a bit simpler in the source. I still think it would be interesting to be able to specify custom sources (i.e. issue description copy-pasted from the source project, etc.), but it could be a future improvement.

tarcieri:
Note: `source` is already an attribute for the provenance of the *package* the advisory is for:

https://docs.rs/rustsec/0.26.2/rustsec/advisory/struct.Metadata.html#structfield.source

It corresponds to the same-named field in Cargo.lock, and defaults to `registry+https://github.com/rust-lang/crates.io-index` if unspecified.

pinkforest:
`origin = ["database", .. ]` then ?

imo `source` field should have been `registry` :woman_shrugging:

tarcieri:
@pinkforest there are other possible sources than a registry.

For example, a company who sources crates via git can maintain an internal vulnerability database with the git URLs of the corresponding packages.

pinkforest:
Yeah that makes sense, even when it says registry+ so it provides the extendable meta format that aligns with the .lock as you say

Anyways can we agree on `origin = ["database", ...]` on this case or any objections to it ?

8573:
What about `advisory-source` or `advisory-origin` or `advisory-database`, to be explicit about what the field is describing?  Having `source` meaning the source of the _package_ and `origin` the origin of the _advisory_ seems unfortunate, when that distinction is not implied by the names "source" and "origin".

(Advisory writers themselves needn't type these field names, right?)

pinkforest:
> (Advisory writers themselves needn't type these field names, right?)

They do. Nonetheless we are not Java and we are not complicating WithLongerFieldThatCouldHaveBeenPerfectlyShortName despite being supposedly more descriptive.  Advisory is already self-evident and I am not sure why it should be repeated in a field.

tarcieri:
Maybe something like `syndicated-from`

amousset:
Updated

amousset:
Updated with:

- [x] Removed license/attribution fields
- [x] A new optional `syndicated-from` field with the recognized `github-advisory-database` value
- [x] A method to get advisory license, computed from advisory origin
- [x] Rendered information for the website

rustsec/src/advisory.rs
```diff
@@ -88,6 +88,15 @@ impl Advisory {
     pub fn withdrawn(&self) -> bool {
         self.metadata.withdrawn.is_some()
     }
+
+    /// License under which this advisory is available, represented as SPDX identifier.
+    pub fn license(&self) -> Option<&str> {
+        match self.metadata.syndicated_from {
+            Some(Origin::GithubAdvisoryDatabase) => Some("CC-BY-4.0"),
+            Some(Origin::Other(_)) => None,
```
amousset, line 96: Not sure what the right interface is in case of unknown origin. Maybe using a `Result` instead?

tarcieri: Personally I think `None` is OK here. The problem is that the license isn't known, not that it's necessarily invalid, and I think `Option` reflects that.

amousset: Or maybe even not provide it, and only render license on the website.

pinkforest: There is no multi-license databases except us after this. Everyone else is CC-BY-4.0.

I agree - let's build features when we need them.

pinkforest:
Two things 

## Long field names with -

Long field names especially with `-` are prone to errors - isn't there a one simple word we can use for this ?

Or even: 
```toml
[context]
field = ["database", ".."]
```

## source field lack of use

Separate issue: https://github.com/rustsec/rustsec/issues/712

rustsec/tests/support/example_advisory_v4_from_ghsa.md
```diff
@@ -0,0 +1,24 @@
+```toml
+id = "RUSTSEC-2001-2101"
+package = "base"
+date = "2001-02-03"
+url = "https://www.youtube.com/watch?v=jQE66WA2s-A"
+categories = ["code-execution", "privilege-escalation"]
+keywords = ["how", "are", "you", "gentlemen"]
+aliases = ["CVE-2001-2101"]
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H"
+syndicated_from = "github-advisory-database"
```
pinkforest, line 10: This should be an array if we use this as we have multiple origins and we should use the well known abbreviations e.g. GHSA, NVD etc. to keep it ergonomic - compact and memorable.

rustsec/src/advisory/metadata.rs
pinkforest: Let's not have aliases - From UX PoV this is very confusing. Long strings are hard to read and prone to errors.

amousset:
I renamed `github-advisory-database` to `GHSA` as it probably be clearer.

About the generic way to set sources for fields I don't think it is worth the complexity, given our current needs.

rustsec/src/advisory/metadata.rs
```diff
@@ -81,4 +81,30 @@ pub struct Metadata {
...
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(from = "String")]
+pub enum Origin {
+    /// This advisory was imported from the GitHub Advisory Database
+    ///
+    /// In this case, it is required to have a GHSA alias.
+    #[serde(rename = "GHSA")]
+    GithubAdvisoryDatabase,
+    /// Other origin.
+    Other(String),
```
tarcieri, line 100: How about actually naming the variant `Ghsa` since that maps directly to the prefix of the original advisory ID?

And as it were, there's already an enum for that, although it doesn't quite meet the requirements needed here: https://docs.rs/rustsec/latest/rustsec/advisory/enum.IdKind.html

pinkforest: It can be case insensitive. All abbreviations are uppercase and enforcing weird capitalisation just creates confusion.

amousset: > And as it were, there's already an enum for that, although it doesn't quite meet the requirements needed here: https://docs.rs/rustsec/latest/rustsec/advisory/enum.IdKind.html

I'm not sure it is semantically the same, there is a difference between the Id issuers and the content source. For example CVE ids are issued by MITRE but a part of the information (cvss scoring, etc.) actually come from nvd (or other sources).

pinkforest:
So I crated a PR in advisory-db https://github.com/rustsec/advisory-db/pull/1507 that has the legal blurb exemption about some imported advisories being licensed optionally as CC-BY 4.0 in individual "exemption" basis

Shnatsel:
Yep, it goes into `database_specific` if there's no field for it in the spec. Kinda weird now that I think about that.

Shnatsel:
I think SPDX license identifiers are the most common way to encode license information in a machine-readable way. The default for RustSec's `toml` format is CC0 but now we're considering adding CC-BY data. I think the CC0 should be called out explicitly in OSV when the license field is absent in TOML.

amousset:
> I think SPDX license identifiers are the most common way to encode license information in a machine-readable way. The default for RustSec's toml format is CC0 but now we're considering adding CC-BY data. I think the CC0 should be called out explicitly in OSV when the license field is absent in TOML.

This is already what's done in this PR (for the rustsec API & rustsec.org, so that's what I has in mind.

amousset:
After some previous discussions I changed the:

```toml
license = "CC-BY-4.0"
attribution = ["I'm the author"]
```

into:

```toml
syndicated_from = "GHSA"
```

Which in retrospect does not seem to me to be the right model as it assumes that _an advisory database = a license+attribution_, which we can't take for granted (as this PR shows!). I think I will go back to the previous format.

amousset:
Nice, I didn't see that they [explicitly accepted](https://docs.github.com/en/site-policy/github-terms/github-terms-for-additional-products-and-features#advisory-database) a link only. So it gives: 

```toml
license = "CC-BY-4.0"
attribution_url = "https://github.com/advisories/GHSA-f8vr-r385-rh5r"
```

amousset:
PR updated. Next identified steps:

* Announce the future addition of CC-BY 4.0 advisories in the database.
* Merge this change (and bump rustsec version as adding the metadata fields is a breaking change)
* After a while, start importing GHSA advisories.

Shnatsel:
Actually, do we even need `attribution_url`? The `url` field is probably already surfaced by UIs, we can simply put the attribution link there and anything else in `references`. I think that's going to make transition the easiest for downstream consumers. What do you think?

amousset:
This might be enough indeed!

amousset:
I'm unsure about this. For example on rustsec.org we would display something like `Advisory available under CC-BY 4.0 license` without an explicit source, which might lead readers to attribute the content to RustSec.

We could do something like "if the advisory has a license requiring attribution, display the url field as attribution url in the license notice", and modify the semantics of the `url` field in the docs to include this.

Shnatsel:
I'm not sure either. @tarcieri @alex any thoughts?

tarcieri:
I think I suggested a field like `syndicated-from` in the past as a way to record attribution.

Is there precedent for this in other vulnerability DBs we could lean on?

amousset:
> Is there precedent for this in other vulnerability DBs we could lean on?

I could not find any, and OSV does not include a field for this information.

> I think I suggested a field like syndicated-from in the past as a way to record attribution.

Indeed, and this was implemented in a previous version of this PR. The way I see the two approaches:

* 1/ Specify advisory origin in the advisory source, and derive license and attribution information from this
  * Pros: simpler to add in advisory source, simpler to document
  * Cons: the correctness of the legal information depends on the tooling, which needs to know about the licensing of an external db. If the terms change in the future, the old logic would provide wrong information.
* 2/ Specify license (and attribution when needed) in the advisory source, and pass it more or less as-is in to our users
  * Pros: advisory source is correct and future-proof, and it is easier to extend to other sources without modifying the tooling
  * Cons: This puts a bit more burden on the advisory authors
 
Currently I think I prefer the second as it simpler and makes less assumptions (but it's not a very strong opinion).

Shnatsel:
I think there are two orthogonal questions here: how do we specify the **origin** and how do we specify the **license**.

Reasonable options for origin that I see are:

1. A new field to specify the origin, just pointing to the DB. Hypothetical name: `syndicated-from`
3. A new field that points to the URL of the exact advisory. Hypothetical name: `source`
4. Put the URL of the original advisory in the existing `url` field

And for the license:

1. Implicit based on the origin (any of the 3 origin options will do)
2. Explicitly specified in a new field

----

I want the format to be **obvious and self-describing.** For the license this means erring on the side of being explicit, so I would specify the license explicitly in a new field. Literally `license = "CC-BY 4.0"`.

I am very much drawn to option 3 for the origin because any reasonable downstream consumers are **automatically compliant,** and no changes to the code are necessary to comply with the new license requirements. By contrast, adding a new field would require existing consumers to manually check it. Although this does not matter for OSV export because it only has `references` array and we could just toss the contents of the new field into it (and/or abuse the `credits` field).

The attribution rule then becomes "if you see `license = "CC-BY 4.0"` field, you must preserve the `url` when displaying advisory data."

The downsides are: 
 - It is not obvious from looking at the format who exactly you are supposed to credit, but that's true of all the options I can think of. We could conceivably add a comment explaining who to credit, but the `toml` crate does not allow writing those and I believe this is not worth the huge hassle it could cause in the tooling.
 - If we ever want to change rustsec data to CC-BY or something else that requires attribution, the "just preserve the URL" rule no longer works, and all consumers would have to take action to accommodate that requirement. But I don't think it is worth forcing everyone to transition *now* because we might *hypothetically* relicense in the future.

---

In light of the above, my proposal is a new `license = "CC-BY 4.0"` field and consumers being required to preserve the contents of the `url` field if it's present. Yay or nay?

8573:
> We could conceivably add a comment explaining who to credit, but the `toml` crate does not allow writing those

Rather than a comment, could it be a(n optional) field?

Shnatsel:
I don't think there is anything stopping us from adding it as a text field.

Subjectively, it feels ugly, and I'd prefer to avoid that for that reason alone.

amousset:
> Reasonable options for origin that I see are:
> 
> 1. A new field to specify the origin, just pointing to the DB. Hypothetical name: syndicated-from
> 2. A new field that points to the URL of the exact advisory. Hypothetical name: source
> 3. Put the URL of the original advisory in the existing url field

I prefer 2. or 3. over 1.

> And for the license:
> 
> 1. Implicit based on the origin (any of the 3 origin options will do)
> 2. Explicitly specified in a new field

I have a strong preference for 2.

So :+1: for `license` + `url`

rustsec/src/advisory/license.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+/// Type representing allowed licenses for advisory content
+enum LicenseVariant {
+    /// CC0-1.0
+    CreativeCommonsZero10,
+    /// CC-BY-4.0
+    CreativeCommonsCc40,
+    /// Other SPDX requirement
+    Other(String),
+}
```
Shnatsel, line 14: Documenting the SPDX identifier to which it serializes is a good idea.

I dread writing out `CreativeCommonsCc40` every time I need it. Can we put the full license name into the comment and an SPDX-like shorthand as the variant name? I'm fine with `CcBy40` if that's what it takes.

rustsec/src/advisory/linter.rs
```diff
@@ -160,7 +162,7 @@ impl Linter {
                             });
                         }
                     }
-                    "url" => {
+                    "url" | "attribution_url" => {
```
Shnatsel, line 165: We decided not to go through with `attribution_url` so this should be dropped.

rustsec/src/osv/advisory.rs
```diff
@@ -150,6 +151,13 @@ pub struct OsvDatabaseSpecific {
     informational: Option<Informational>,
 }
 
+#[derive(Debug, Clone, Serialize)]
+pub struct MainOsvDatabaseSpecific {
+    license: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    attribution_url: Option<Url>,
```
Shnatsel, line 158: I'd prefer not to have a separate `attribution_url` field in OSV. 

Since this is a database-specific field, licence compliance would require special handling for RustSec in case the attribution URL is not also duplicated in the `references` field.

I'd rather put the attribution URL as the first link in references. Same logic as the `url` field - all reasonable users are automatically compliant.

rustsec/tests/support/example_advisory_v4_from_ghsa.md
```diff
@@ -0,0 +1,25 @@
...
+
+[affected]
+arch = ["x86"]
+os = ["windows"]
+functions = { "base::belongs::All" = ["< 1.2.3"] }
+```
+
+# All your base are belong to us
+
+You have no chance to survive. Make your time.
```
Shnatsel, line 25: :rocket: :earth_asia:

rustsec/tests/support/example_advisory_v4_from_ghsa.md
```diff
@@ -0,0 +1,25 @@
+```toml
+id = "RUSTSEC-2001-2101"
+package = "base"
+date = "2001-02-03"
+url = "https://www.youtube.com/watch?v=jQE66WA2s-A"
```
Shnatsel, line 5: This should be moved to `references` and the URL should point to GHSA

rustsec/src/osv/advisory.rs
amousset: This was because of the field in the advisory metadata, once we decide to use `url` it makes no sense anymore.

rustsec/src/advisory/license.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+/// Opaque type representing allowed licenses for advisory content
+pub struct License {
+    inner: LicenseVariant,
+}
```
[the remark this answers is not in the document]
amousset: The idea was (IIRC) to store a enum to allow easier usage internally but to only expose the SPDX identifier as an `str` to keep the API simple and allow future evolution. As for why using manual `Deserialize`/`Serialize`, no idea.

This is not linked to serialization.

rustsec/src/advisory/license.rs
amousset: Works for me.

rustsec/src/advisory/license.rs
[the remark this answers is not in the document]
Shnatsel: I think you could achieve the exact same thing simply by making the enum `#[non_exhaustive]`? That would make for a much smaller API surface.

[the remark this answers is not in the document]
amousset: Well that introduces a new type in the API, but I'm not opposed to it.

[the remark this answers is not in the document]
amousset: (And that makes the implementation a lot simpler indeed)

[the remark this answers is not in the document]
Shnatsel: We're breaking the API in the next release anyway, so might as well add a type.

Shnatsel:
Looks good to go. Thanks!

We'll also need to update `cargo audit` to surface these URLs when the license is CC-BY.

PR #688, opened 2022-09-09
Recursive directory scanning for `cargo audit bin`

Shnatsel:
Initial, WIP support for scanning a directory recursively, rather like Trivy.

https://github.com/rustsec/rustsec/pull/635 should land first.

Other than that, there is work remaining to clean this up - error handling, etc.


Shnatsel:
I'm going to go ahead and close this. We provide a recipe for recursive scanning in the README.md using `fd`, and reimplementing this inside cargo-audit would require the entire complexity of `fd` in the long run. We have enough complexity as it is.

PR #702, opened 2022-09-27
Clarify location of audit config

threema-danilo:
It took me a while to realize that the audit config can be located in the project directory as well. This PR adds some documentation for that.


tarcieri:
Thanks!

PR #710, opened 2022-10-02
I/O operation failed: failed to read directory `rustsec/crates/.DS_Store`

kornelski:
macOS has an awful habit of dropping `.DS_Store` files everywhere, which breaks scanning of the advisory-db.

### DIFF
--- a/rustsec/src/database.rs
+++ b/rustsec/src/database.rs
@@ -50,8 +50,18 @@ impl Database {
 
             if let Ok(collection_entry) = fs::read_dir(&collection_path) {
                 for dir_entry in collection_entry {
-                    for advisory_entry in fs::read_dir(dir_entry?.path())? {
-                        advisory_paths.push(advisory_entry?.path().to_owned());
+                    let dir_entry = dir_entry?;
+                    if !dir_entry.file_type()?.is_dir() {
+                        continue;
+                    }
+                    for advisory_entry in fs::read_dir(dir_entry.path())? {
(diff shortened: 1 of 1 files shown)


rustsec/src/database.rs
```diff
@@ -50,8 +50,18 @@ impl Database {
...
-                    for advisory_entry in fs::read_dir(dir_entry?.path())? {
-                        advisory_paths.push(advisory_entry?.path().to_owned());
+                    let dir_entry = dir_entry?;
+                    if !dir_entry.file_type()?.is_dir() {
+                        continue;
+                    }
+                    for advisory_entry in fs::read_dir(dir_entry.path())? {
+                        let advisory_path = advisory_entry?.path();
+                        let file_name = advisory_path.file_name().and_then(|f| f.to_str());
+                        // skip dotfiles
```
Shnatsel, line 60: Could you also put the comment about the specific MacOS files here, so that it's clear *why* we skip dotfiles?

Shnatsel (approved):
One nit, otherwise looks good to me. Thanks!

Shnatsel:
I think CI just timed out, rerunning it should fix it automatically.

Shnatsel:
Merged, thanks!

How critical is this? Should I push a point release right now, or wait till the next release and roll it up with other changes?

kornelski:
I can keep deleting the .DS_Store files for a while longer.

PR #721, opened 2022-10-07
Revert "Don't use --locked in release workflow to allow publishing again"

Shnatsel:
#718 re-added Cargo.lock and somehow CI didn't fall over, so re-enable --locked publishing.

This reverts commit e8d343beefbd49d986b51fb552984f25d12a47bd.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,7 +93,7 @@ jobs:
         with:
           use-cross: ${{ matrix.cross }}
           command: build
-          args: --release --features "${{ join(matrix.features, ',') }}" --target ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


tarcieri:
@Shnatsel I believe the issue was `libgit2-sys` is using namespaced and/or weakly activated features, which have a rather confusing failure mode and error message on Rust versions prior to v1.60, which added support.

The workaround is to compute Cargo.lock using the MSRV release of cargo.

Note that the presence of these features prevents an entire release from being parsed, so even the presence of `rust-version` doesn't help.

Shnatsel:
Oh, that's what caused all those errors :facepalm: 

Thanks for explaining this!

PR #729, opened 2022-10-23
`cargo audit bin` without `cargo auditable` based on panic messages

### LINKED ISSUE #724: Best-effort dependency detection in binaries built without `cargo auditable`
The panic messages contain paths to dependencies, which (when built with Cargo) include the crate name and version. This would let us detect *some* dependencies even in black-box binaries when running `cargo audit bin`.

A quick proof-of-concept implementation looks like this:

```rust
use std::collections::BTreeSet;
use regex::bytes::Regex;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let path = std::env::args_os().nth(1).unwrap();
    let file = std::fs::read(path)?;
    let re =
        Regex::new("(?-u)cargo/registry/src/[^/]+/(?P<crate>[0-9A-Za-z_-]+)-(?P<version>[0-9A-Za-z-+.]+)/")
            .unwrap();
    let mut versions: BTreeSet<(&str, &str)> = BTreeSet::new();
    for c in re.captures_iter(&file) {
        versions.insert((
            std::str::from_utf8(c.name("crate").unwrap().as_bytes())?,
            std::str::from_utf8(c.name("version").unwrap().as_bytes())?,
        ));
    }
    for (krate, version) in versions.iter() {
        println!("{krate} {version}")
    }
    Ok(())
}
```

### LINKED ISSUE #735: `--deny=warnings` results in incomplete report when scanning multiple binaries
`cargo audit bin --deny=warnings file1 file2` exits immediately and does not scan `file2` if `file1` has warnings.

This is due to the following line of code:

https://github.com/rustsec/rustsec/blob/a7310a402b2809a88d62cdd770a6401fae2382ae/cargo-audit/src/presenter.rs#L178-L181

Instead of exiting immediately, this should at least be reported back from the audit function and then acted upon by the caller.

Shnatsel:
Closes #724
Fixes #735

Adds tests for exit code on unmaintained crates with and without `--deny=warnings`

TODO:
 - [x] Print a warning when falling back to dependency detection based on panic messages
 - [ ] ~~Add a flag to require presence of audit info and treat its absence as an error~~ will be addressed in the future
 - [x] Add tests
 - [x] Update help text to clarify that building with `cargo auditable` is no longer required
 - [x] Update README.md to clarify that building with `cargo auditable` is no longer required
 - [x] Do something about the output not specifying what file it refers to if `--quiet` is used


amousset:
From a UX standpoint, I think the fallback on panic messages shouldn't be totally transparent, as it may give a false sense of confidence.

IMO we should have an info or warning message when falling back to panic messages, and we also probably need a cli flag to disable it (e.g. for users who would build everything with cargo-auditable and would consider the absence of deps metadata an audit failure).

Shnatsel:
Yeah, I totally agree about printing a warning. I was meaning to do that.

Adding a flag that requires audit info is a good idea, I'll add it to the TODO.

Shnatsel:
Ready for review!

I'm postponing enabling/disabling panic-based dependency detection from the CLI. This PR is starting to incorporate unrelated fixes, so I'd rather merge it and add any missing features later than roll up fixes and feature work into a single PR.

PR #741, opened 2022-10-30
Add cargo-auditable to home page

amousset:


amousset:
Then we can probably publish the new home page.

Shnatsel:
Good idea, thanks!

PR #751, opened 2022-11-05
Consolidate CODE_OF_CONDUCT.md files into one; switch to Rust code of Conduct

Shnatsel:
@tarcieri is having many CODE_OF_CONDUCT.md files still relevant, or is it just a historical quirk?


tarcieri (approved):
I think it's a leftover vestige from before the various crates were combined into a monorepo. At any rate, it's fine to remove them.

Issue #686, opened 2022-09-07
Add fields `published` and / or `serial`


pinkforest:
**Two reasons**

One is that the presentation layer is naively using the `date` field which reasonable person would expect to be when the advisory was published however it is the date when the underlying issue was raised with the vendor / maintainer making it hard to change that field.

Second is that we often update advisories and we should have a serial number for those updates e.g. acting like a DNS SOA Serial.

These fields would be handled by the automation automatically so the curators wouldn't need to touch them.

`published = YYYY-MM-DD`

`updated = U` where the `U` is the serial number of the advisory itself starting from 0 from the initial published advisory

Before we have these fields the website should probably generate the initial published / updated: https://github.com/rustsec/rustsec/issues/685

tarcieri:
This information is already available via the git history. Can't we use that rather than adding manual annotations?

Likewise the RUSTSEC ID already provides a serial number of the order in which advisories are published.

brson:
Ordering the website list by ID would solve the problem I had related to https://github.com/rustsec/rustsec/issues/685, where newly published advisories appeared far down the page where I wasn't finding them.

pinkforest:
It wouldn't be manual annotation as assign-id's would do that field

Serial numbers are by year and I wouldn't know whether 2021-0123 comes earlier than 2022-0004 without looking at commit history on which you need the git history

Issue #685, opened 2022-09-07
Website is using the `date` field for sorting


pinkforest:
Date field can be backdated and the advisories appear out-of-order in https://rustsec.org/advisories/

Didn't the website have access to the repository and it could just look at the id-assigned commit date(s) until we get new fields to denote when it was published / updated ?

Yeah I know it requires index / caching and we can update those fields based on that later but before that ? :woman_shrugging: 

Simplest perhaps is probably first order by year and then order by serial on year instead of using date to sort ?

And then to generate index advisory -> date published from commit log for the website use

brson:
Sorting the website listing by ID would be more intuitive to me. When I collect rustsec advisories for summarization I look at the order they are listed on the website, and I hadn't realize that some new ones were listed far down the page.

amousset:
It has been deployed.

PR #754, opened 2022-11-05
Sort advisories by publication date

amousset:
Tries to solve the ordering problem (ref. #685, #686) for the website and the atom feed.

An advisory has 3 dates:

* Reporting date, present in its metadata. This date is used for determining the year in the RustSec ID.
* Publication date, derived from the advisory file's creation date in git.
* Modification date, derived from the advisory file's modification date in git.

This changes the sorting order to use publication date on the website index and atom feed, and display it in the list.

image

As a consequence, the advisories in the index are not grouped by advisories year, so the grouping is removed (but seems to be way less relevant anyway as the number of advisories by year grows).

The three dates are now present in advisory details.
image

### DIFF
--- a/admin/src/commands/lint.rs
+++ b/admin/src/commands/lint.rs
@@ -30,7 +30,7 @@ impl Runnable for LintCmd {
             _ => unreachable!(),
         };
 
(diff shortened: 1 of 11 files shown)


pinkforest (approved):
sgtm - 

Would be also nice to be able to switch sort between modified / created and between desc / asc

PR #769, opened 2022-12-10
Add client-side glob expansion on Windows

Shnatsel:
I've bumped version to 0.18.0 because this is technically a breaking change - glob paths now have to be escaped on Windows, which wasn't a requirement previously. 

But I'm not very familiar with argument parsing on Windows so I may be mistaken about the impact of this. In which case erring on the side of caution is prudent.

### DIFF
--- a/admin/src/bin/rustsec-admin/main.rs
+++ b/admin/src/bin/rustsec-admin/main.rs
@@ -7,5 +7,11 @@ use rustsec_admin::application::APPLICATION;
 
 /// Boot the `rustsec-admin` CLI application
 fn main() {
-    abscissa_core::boot(&APPLICATION);
+    // copied from `abscissa` code:
+    // https://github.com/iqlusioninc/abscissa/blob/7c9ab7976145d2920a99ef00eba845efab3247a1/core/src/application.rs#L171-L177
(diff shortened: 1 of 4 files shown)


admin/src/bin/rustsec-admin/main.rs
```diff
@@ -7,5 +7,11 @@ use rustsec_admin::application::APPLICATION;
 
 /// Boot the `rustsec-admin` CLI application
 fn main() {
-    abscissa_core::boot(&APPLICATION);
+    // copied from `abscissa` code:
+    // https://github.com/iqlusioninc/abscissa/blob/7c9ab7976145d2920a99ef00eba845efab3247a1/core/src/application.rs#L171-L177
+    // so that we could inject wild::args() instead of std::env::args()
+    // which provides glob expansion on Windows
+    let args = std::env::args();
```
tarcieri, line 14: Seems like you didn't actually use `wild::args`?

Shnatsel:
I'm having second thoughts about this being the right approach, and about glib expansion on Windows in the first place. Maybe it is a reasonable default, but then we'd have to provide a command-line flag to disable it, and that cannot be done at this level of abstraction. I'm just going to close this for now.

PR #776, opened 2022-12-30
cargo-lock: implement From<Name> for String

00xc:
Since `Name` already stores a `String` internally, this allows for a cheap conversion. Implementing this trait also implicitly adds `Into<String> for Name` - this is more desirable than calling `name.to_string()` (which implies an allocation) if one wants an owned `String` and can consume the `Name` anyway.


tarcieri:
Thanks!

Issue #795, opened 2023-01-24
cargo-audit: macOS build failures in CI when testing `auditable`


tarcieri:
https://github.com/rustsec/rustsec/actions/runs/3992756517/jobs/6848877427

```
failures:

---- auditable_binary_without_vulnerabilities_passes stdout ----
thread 'auditable_binary_without_vulnerabilities_passes' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', cargo-audit/tests/binary_scanning.rs:66:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- auditable_binary_with_vulnerabilities_fails stdout ----
thread 'auditable_binary_with_vulnerabilities_fails' panicked at 'poisoned cmd mutex!: PoisonError { .. }', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/abscissa_core-0.6.0/src/testing/runner.rs:195:39

---- panicking_binary_with_vulnerabilities_fails stdout ----
thread 'panicking_binary_with_vulnerabilities_fails' panicked at 'poisoned cmd mutex!: PoisonError { .. }', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/abscissa_core-0.6.0/src/testing/runner.rs:195:39

---- panicking_binary_without_vulnerabilities_passes stdout ----
thread 'panicking_binary_without_vulnerabilities_passes' panicked at 'poisoned cmd mutex!: PoisonError { .. }', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/abscissa_core-0.6.0/src/testing/runner.rs:195:39


failures:
    auditable_binary_with_vulnerabilities_fails
    auditable_binary_without_vulnerabilities_passes
    panicking_binary_with_vulnerabilities_fails
    panicking_binary_without_vulnerabilities_passes
```

tarcieri:
Unless there's a quick fix, I'd suggest disabling the `--all-features` test on macOS for now so it's not blocking other development

tarcieri:
I opened #796 which makes various changes to the CI config, including re-enabling the Windows build

Shnatsel:
Riiiight, of course the proprietary platform I don't have access to will fail :facepalm: 

Time to take up that offer from a Mac renting company and put an ad for them in `cargo-auditable`'s README, I suppose.

Shnatsel:
[...]

PR #796, opened 2023-01-24
cargo-audit: disable macOS CI for `--all-features`

tarcieri:
It's currently failing. See #795.

This also attempts to re-enable the Windows build.


tarcieri:
It might be worth splitting each OS into a separate config to enable more flexibility, given how Windows and macOS both have (different) issues

PR #801, opened 2023-01-25
rustsec: bump `cargo-edit` to v0.11

tarcieri:


jonhoo:
This also blocks updating the `git2` (now at 0.16) and `crates-index` (now at 0.19) dependencies:

https://github.com/rustsec/rustsec/blob/1084941e81632b08b08dc234f36fd818ca3db1af/rustsec/Cargo.toml#L28-L29

The git2 0.16 bump (like the 0.15 bump) is [mostly backwards compatible](https://github.com/rust-lang/git2-rs/blob/master/CHANGELOG.md), so we _could_ just expand the range to `>=0.15, <0.17` (like we did in https://github.com/rustsec/rustsec/pull/662), but given that `crates-index 0.19` [takes a `git2 = "0.16"` dependency](https://crates.io/crates/crates-index/0.19.0/dependencies), there wouldn't be an advantage to doing so as Cargo would always be forced to pick 0.16 anyway. 0.18.11 [does permit](https://github.com/frewsxcv/rust-crates-index/commit/415875ccfa052bf07a7641db1e5fecd820e20c5f) all the versions of `git2`, so you could bump `git2` first and then `crates-index` if you'd prefer to split them up.

The `crates-index` bump [is just for a bump of `toml` and `git2`](https://github.com/frewsxcv/rust-crates-index/blob/master/README.md#migration-from-018), so should be an easy upgrade.

tarcieri:
@jonhoo bumping `git2` is blocked on `cargo-edit` updating from v0.14: https://github.com/killercup/cargo-edit/issues/770

tarcieri:
Also, this PR is blocked on [finding a replacement for `cargo_edit::LocalManfiest::upgrade`](https://docs.rs/cargo-edit/0.9.1/cargo_edit/struct.LocalManifest.html#method.upgrade)

jonhoo:
Ah, great, thanks for the breadcrumb trail!

jonhoo:
Looks like `cargo-edit` bumped `git2` yesterday in https://github.com/killercup/cargo-edit/pull/833, it just hasn't been released to crates.io yet.

tarcieri:
Cool, hopefully there will be a release soon which will unblock the other updates

jonhoo:
Speak of the devil. 0.11.8 is out: https://crates.io/crates/cargo-edit/0.11.8

tarcieri:
Cool, so to get this over the line we just need to figure out a replacement for `LocalManifest::upgrade` as [mentioned above](https://github.com/rustsec/rustsec/pull/801#issuecomment-1405584311)

jonhoo:
If the [implementation of it](https://docs.rs/cargo-edit/0.9.1/src/cargo_edit/manifest.rs.html#266-292) just uses publicly-available APIs, I guess you could just copy-paste that code from the old version?

tarcieri:
I opened an issue here: https://github.com/killercup/cargo-edit/issues/837

jonhoo:
So I guess the way to go (at least in the short term) then is to copy-paste in both the `upgrade` method and any other internal method dependencies. Long-term, it's to move away from depending on the library API of `cargo-edit`.

pinkforest:
I forked cargo-edit to cargo-edit-9 - we may just use that temporarily I guess
PR here:
- https://github.com/rustsec/rustsec/pull/831

Issue #807, opened 2023-02-07
Lacking regular crates.io releases


kornelski:
The last release of rustsec on crates.io is from November last year, and it has now-outdated dependencies. This causes duplicate crates in projects depending on rustsec.

Could you release rustsec crate more often?

tarcieri:
As you've probably discovered in #808 we're dealing with some rather unfortunate dependency problems surrounding `cargo-edit` and `git2`. It sure would be nice if the upgrade was that easy, but it isn't.

You might want to look through the PRs and issues on this, namely #801. We can't easily upgrade `cargo-edit` without removing the `--fix` functionality as the relevant functionality in `cargo-edit` has been removed and it can no longer be used for our purposes.

We will likely need to vendor the relevant functionality into the `rustsec` crate, which I've been meaning to do but have been busy with other things.

tarcieri:
The real blocker here is #801, so I'm going to close this.

Shnatsel:
Thanks to @pinkforest the next release is unblocked, so I published v0.26.5 just now. If you encounter any more issues with outdated dependencies, please let us know.

PR #808, opened 2023-02-07
Upgrade git2 dependency

kornelski:
In `crates-index` I had to bump `git2` to the specific latest version, because I had git2 types are in the public API, and leaving it flexible didn't work as well as I'd hoped — cargo picks the latest per crate anyway, even if it doesn't match versions used in other crates.

### DIFF
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         toolchain:
-          - 1.60.0 # MSRV
+          - 1.64.0 # MSRV
           - stable
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


tarcieri:
As I mentioned in #807, this doesn't work until #801 is resolved, because `cargo-edit` v0.11 no longer has the functionality we use.

See also: https://github.com/killercup/cargo-edit/issues/837

kornelski:
That's unfortunate. I'm using `rustsec` only for the database access, and never interact with individual crates, so the `--fix` option is irrelevant for my usage.

Maybe you could split rustsec into rustsec-cli and rustsec-db?

tarcieri:
`cargo-audit` is the `rustsec` CLI. `rustsec` is a library

Issue #812, opened 2023-02-10
`cargo audit bin` reports Windows-only issues in Linux binaries


Shnatsel:
For example, https://rustsec.org/advisories/RUSTSEC-2023-0001 is flagged in Linux binaries when using `cargo audit bin`.

It may not be possible to do proper filtering if there's e.g. only BSD systems affected, but it should be possible to discard Windows-only vulnerabilities in ELF and Mach-O, for example.

PR #814, opened 2023-02-14
`cargo audit bin`: filter advisories by binary type

Shnatsel:
Fixes https://github.com/rustsec/rustsec/issues/812

But with a pretty significant limitation of not filtering warnings, because the `rustsec` crate simply does not expose the `affected` field on warnings. So this will probably require an API break in `rustsec` to fix, one way or another.

TODO:

- [x] Figure out how to make `binfarce` an optional dependency?
- [ ] Add tests


amousset (approved):

amousset:
Everything that lowers the amount of false positives is good to have :+1:

Shnatsel:
Thanks! In that case I'll add tests and merge this. 

I'll open another issue to track filtering of the warnings.

Shnatsel:
Platform-specific advisories convenient for testing this:

https://rustsec.org/advisories/RUSTSEC-2023-0001

https://rustsec.org/advisories/RUSTSEC-2021-0119

https://rustsec.org/advisories/RUSTSEC-2022-0049.html

Shnatsel:
I'm going to merge this to avoid future editing conflicts. I'll work on tests in a separate PR because they require large changes to the test infrastructure, which may be more controversial than this PR and require some iteration.

PR #815, opened 2023-02-20
Re-enable MacOS CI with `--all-features`

Shnatsel:
The issue is not reproducible on a local Mac so re-enabling to check if it was transient


Shnatsel:
Addresses #795

PR #817, opened 2023-02-22
Linter: Clean up crates.io package name check error

### LINKED ISSUE #622: Linter: Clean up crates.io package name check error
Noticed another error as part of fixing #619 via #620

e.g. when a crate name is missing from crates.io we get:
`error: error linting advisory DB ../../: crates.io index error: crates.io package name does not match package name in advisory for rustdecimal`

What is particular about this error is that it does not show which crate causes that and instead hals at middle

It should error same as the others e.g. if advisory toml has invalid keys etc. where it shows which exact advisory caused the error

thewizzy:
Added the advisory id to the output to help identify the advisory with the error.

Resolves #622

### DIFF
--- a/admin/src/linter.rs
+++ b/admin/src/linter.rs
@@ -134,8 +134,9 @@ impl Linter {
 
             fail!(
                 ErrorKind::CratesIo,
-                "crates.io package name does not match package name in advisory for {}",
-                advisory.metadata.package.as_str()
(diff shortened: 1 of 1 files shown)


Shnatsel:
Looks good, thanks! And sorry for the review delay.

Shnatsel (approved):

PR #839, opened 2023-03-22
Do not run all tests from the default feature set twice

Shnatsel:
The only non-default feature is `fix`


Shnatsel:
@tarcieri I think the test matrix can be re-unified now that all tests are passing. There's a TODO for it in your name so I assume you know how to do that?

Shnatsel:
I'm going to merge this to speed up the worfklow in the short term, but I'd be happy to see this cleaned up in follow-up PRs.

PR #841, opened 2023-03-23
feature: render third-party crates json as API

Folyd:
This PR aimed to provide an Open API for vulnerability advisories of third-party crates.

The approach is to render each crate as a JSON file:

```
$ tree _site/api
_site/api
├── abi_stable.json
├── abomonation.json
├── abox.json
├── acc_reader.json
├── actix_codec.json
├── actix_http.json
├── actix_service.json
├── actix_utils.json
├── actix_web.json
....
├── ws.json
├── xcb.json
├── xml_rs.json
├── yaml_rust.json
├── yottadb.json
└── zeroize_derive.json

0 directories, 409 files
```

The user can use `curl` command to fetch the vulnerability info about the given crate.
The API format is https://rustsec.org/api/{crate_name}.json, for example:
`curl https://rustsec.org/api/actix_web.json` will get the json response :

```json
[
  {
    "advisory": {
      "id": "RUSTSEC-2018-0019",
      "package": "actix-web",
      "title": "Multiple memory safety issues",
      "description": "Affected versions contain multiple memory safety issues, such as:\n\n - Unsoundly coercing immutable references to mutable references\n - Unsoundly extending lifetimes of strings\n - Adding the `Send` marker trait to objects that cannot be safely sent between threads\n\nThis may result in a variety of memory corruption scenarios, most likely use-after-free.\n \nA significant refactoring effort has been conducted to resolve these issues.",
      "date": "2018-06-08",
      "aliases": [],
      "related": [],
      "collection": "crates",
      "categories": [
        "memory-corruption"
      ],
      "keywords": [],
      "cvss": null,
      "informational": null,
      "references": [],
      "source": null,
      "url": "https://github.com/actix/actix-web/issues/289",
      "withdrawn": null
    },
    "affected": null,
    "versions": {
      "patched": [
        ">=0.7.15"
      ],
      "unaffected": []
    }
  }
]
```

### DIFF
--- a/admin/src/web.rs
+++ b/admin/src/web.rs
@@ -10,10 +10,14 @@ use chrono::{Duration, NaiveDate, Utc};
 use comrak::{markdown_to_html, ComrakOptions};
 use rust_embed::RustEmbed;
 use rustsec::advisory::Id;
+use rustsec::database::Query;
 use rustsec::repository::git::GitModificationTimes;
 use rustsec::repository::git::GitPath;
+use rustsec::Collection;
+use rustsec::Database;
 use rustsec::{advisory, Repository};
 use std::collections::{HashMap, HashSet};
+use std::io::Write;
 use std::str::FromStr;
 use std::{
     fs::{self, File},
@@ -111,6 +115,11 @@ pub fn render_advisories(output_folder: PathBuf) {
 
     // Get advisories
     let db = rustsec::Database::fetch().unwrap();
+
+    // Render api json files
+    render_api(&db, &output_folder);
+    status_ok!("Completed", "API rendered");
+
     let mut advisories: Vec<(rustsec::Advisory, advisory::Date, advisory::Date)> = db
         .into_iter()
         .map(|a| {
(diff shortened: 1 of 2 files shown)


Shnatsel:
I'm not sure RustSec should provide an API in a custom format without tooling support. While this looks simple on the surface, matching the version specification correctly is non-trivial. The `rustsec` crate will be required to implement it correctly (standalone solutions such as the `semver` crate has issues with edge cases such as pre-releases).

[OSV](https://osv.dev/) is a simpler format to consume, and we already [export](https://github.com/rustsec/advisory-db/tree/osv) all our data to OSV format. But if we're using OSV then there's no point in a custom API because it's just going to be a worse version of the already existing very convenient [OSV API](https://osv.dev/#use-the-api) which does all the version matching for you and just returns the advisories, if any.

If you have use cases not served by the `rustsec` crate and the OSV API, please let us know. But as it stands I recommend using either of those things instead of building a custom API and making the API consumer deal with version matching.

Folyd:
Wow, good to know the OSV API, that is great! So I'll close this PR, thanks.

Issue #750, opened 2022-11-03
Get rid of OpenSSL


Shnatsel:
`cargo audit` currently depends on OpenSSL:

```
cargo tree --all-features -i openssl-sys
openssl-sys v0.9.77
├── git2 v0.14.2
│   ├── cargo-edit v0.9.1
│   │   └── rustsec v0.26.3 (/home/shnatsel/Code/rustsec/rustsec)
│   │       └── cargo-audit v0.17.3 (/home/shnatsel/Code/rustsec/cargo-audit)
│   ├── crates-index v0.18.10
│   │   ├── cargo-edit v0.9.1 (*)
│   │   └── rustsec v0.26.3 (/home/shnatsel/Code/rustsec/rustsec) (*)
│   └── rustsec v0.26.3 (/home/shnatsel/Code/rustsec/rustsec) (*)
├── libgit2-sys v0.13.2+1.4.2
│   └── git2 v0.14.2 (*)
├── libssh2-sys v0.2.23
│   └── libgit2-sys v0.13.2+1.4.2 (*)
├── native-tls v0.2.10
│   └── ureq v2.5.0
│       └── cargo-edit v0.9.1 (*)
└── openssl v0.10.42
    └── native-tls v0.2.10 (*)
```

It's ironic to depend on a large source of CVEs in a tool that looks for CVEs.

#923 and #925 made progress in transitioning from `libgit2` to `gix`, and it is now gone in most configurations. There are some users of `libgit2` still remaining:

- [ ] `rustsec-admin` crate uses it via `crates-index`, needs to be converted to `tame-index`.
- [ ] `cargo-edit-9` uses `crates-index` and `libgit2` directly. This is used only when the `fix` feature is enabled. We should rewrite it outright, see https://github.com/rustsec/rustsec/issues/938

tarcieri:
I would suggest removing the `git` dependency entirely and going the `ureq` route. There's so much environmental for errata that has lead to a litany of tiny issues for people.

The database tarball located at:

https://github.com/rustsec/advisory-db/archive/main.tar.gz

...is ~150kB, so we can just download and unpack that to do updates, caching etags or what not to determine if an update is needed.

Shnatsel:
https://github.com/henry40408/git2-hyper provides a `hyper`-based transport for `libgit2` and exposes `rustls` as an option.

Byron:
[...]

PR #843, opened 2023-03-24
Download the database over HTTPS

Shnatsel:
Part of https://github.com/rustsec/rustsec/issues/750

Not actually wired up to `cargo audit` yet, just shows what the approach might look like


rustsec/src/database.rs
```diff
@@ -15,10 +15,13 @@ use crate::{
...
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
-#[cfg(feature = "git")]
-use crate::repository::git;
+/// Directory under `~/.cargo` where the advisory-db repo will be kept
+const ADVISORY_DB_DIRECTORY: &str = "advisory-db";
+
+/// The default URL from which the DB is downloaded
+const DEFAULT_DB_URL: &str = "https://github.com/zip-rs/zip/archive/refs/heads/master.zip";
```
tarcieri, line 24: This seems bogus in multiple ways... just for testing I guess?

FWIW the tarball seems a lot smaller:

- 463 kB: https://github.com/rustsec/advisory-db/archive/refs/heads/main.zip
- 180 kB: https://github.com/rustsec/advisory-db/tarball/main

Shnatsel: Anything else I should fix here other than switching from .zip to .tar.gz?

tarcieri: Looking good so far otherwise. Hopefully etags work for those resources.

Shnatsel: Well they do provide an etag in the response, so the worst case is we don't rely on the 304 and instead check the etag in the response headers and stop there. So no worries there.

PR #848, opened 2023-03-27
Fix #736 - Cargo audit self advisories repeated

### LINKED ISSUE #736: Advisories for `cargo audit` itself are printed for every scanned binary (instead of once)
`cargo_audit::presenter::print_report` contains the following code:

https://github.com/rustsec/rustsec/blob/a7310a402b2809a88d62cdd770a6401fae2382ae/cargo-audit/src/presenter.rs#L166-L176

Since this function is run for every binary scanned, it will also print out the self-scan data for every single binary. This is unnecessarily noisy; this report should only be printed once, at the end.

thewizzy:
`audit()` was called by both the binary, multi binary and lockfile paths. This included a call to check for self advisories against cargo-audit.
For multi-binary this no only printed the results of self scanning multiple times but ran the checks multiple times.

Now the lockfile and binary execution paths check for self issues and print the results.

I have also combined the two older checks to a single check and message.

Tests pass but I believe there are no tests covering this.

### DIFF
--- a/cargo-audit/src/auditor.rs
+++ b/cargo-audit/src/auditor.rs
@@ -146,7 +146,14 @@ impl Auditor {
 
         self.presenter.before_report(lockfile_path, &lockfile);
 
-        self.audit(&lockfile, None)
+        let report = self.audit(&lockfile, None);
+
+        let self_advisories = self.self_advisories();
+
+        self.presenter
+            .print_self_report(self_advisories.as_slice());
+
+        report
     }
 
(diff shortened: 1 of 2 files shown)


Shnatsel:
Fixes #736

thewizzy:
> Fixes #736

Sorry, i thought it was linked in the Issue but it was just a reference! 
Thank you.

Shnatsel:
Looks good. Thank you!

PR #854, opened 2023-03-30
Link to rustsec/audit-check

SergioGasquez:
Since action-rs is not maintained, I'd point to your updated version of the action.


tarcieri:
Thanks!

PR #857, opened 2023-04-03
Re-unify CI definition into a test matrix

Shnatsel:


Shnatsel:
Well now windows 1.60.0 target times out even thought it's the exact same invocation and I have no clue why

Issue #696, opened 2022-09-19
`cargo audit` gets local crates confused with remote crates of the same name


banool:
## Repro
Pull this repo:
```
git clone https://github.com/aptos-labs/aptos-core.git
git checkout 63f1890
```

Run cargo audit. You'll see that it reports the following problem:
```
Crate:     scratchpad
Version:   0.1.0
Title:     move_elements can double-free objects on panic
Date:      2021-02-18
ID:        RUSTSEC-2021-0030
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0030
Solution:  Upgrade to >=1.3.1
Dependency tree:
scratchpad 0.1.0
├── vm-validator 0.1.0

---

etc.
```

The problem is we're not actually using the remote `scratchpad`, this is a completely different crate in the local workspace:
```
$ cargo tree -i scratchpad
scratchpad v0.1.0 (/Users/dport/a/core/storage/scratchpad)
```

You can confirm this by looking at the lockfile, nothing depends on the `scratchpad` crate from crates.io.

tarcieri:
Can you confirm what version of `cargo audit` you’re using?

This problem should’ve been fixed by https://github.com/RustSec/rustsec/pull/541 which was released in `cargo-audit` v0.17

banool:
Seems like I'm using that version:
```
$ cargo-audit
cargo-audit 0.17.0
Tony Arcieri <bascule@gmail.com>
Audit Cargo.lock for crates with security vulnerabilities

USAGE:
    cargo-audit [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -v, --verbose    Increase verbosity
    -V, --version    Print version information

SUBCOMMANDS:
    audit    Audit Cargo.lock files for vulnerable crates
    help     Print this message or the help of the given subcommand(s)
```

tarcieri:
It looks like [`rustsec::report::Settings::query`](https://docs.rs/rustsec/latest/rustsec/report/struct.Settings.html#method.query) isn't setting the `SourceId` in the generated query.

Shnatsel:
There's a minimal reproduction in #821

ElBe-Plaq:
Still possible in `0.18.3` (different repository).

Shnatsel:
Looking at the code, I think we'll need to change how it works and make breaking change to the API to fix this.

[...]

PR #859, opened 2023-04-04
Fix `is_default_registry` for sparse index

Jake-Shadle:
`SourceId::is_default_registry` would incorrectly report that the sparse crates.io index was not crates.io.

### DIFF
--- a/cargo-lock/src/package/source.rs
+++ b/cargo-lock/src/package/source.rs
@@ -17,6 +17,8 @@ use std::path::Path;
 
 /// Location of the crates.io index
 pub const CRATES_IO_INDEX: &str = "https://github.com/rust-lang/crates.io-index";
+/// Location of the crates.io sparse HTTP index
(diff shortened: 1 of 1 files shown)


Shnatsel:
Oh, we'll need to ship a fix soon. Without this PR `cargo audit` would have false negatives for anyone using the sparse index.

tarcieri:
@Shnatsel I don't think that's true. This method isn't used in vulnerability auditing, the logic to parse the sparse registry entries doesn't exist in current releases, and it seems like the sources aren't being considered in the queries anyway per #696

Shnatsel:
Do I understand correctly that we would need to modify the `cargo-lock` and `rustsec` crates to support auditing crates coming from a sparse registry?

tarcieri:
No, that is not the case because source-based filtering isn't working anyway.

To resolve #696 correctly we will need to do those things.

PR #862, opened 2023-04-11
Feature: filter returned vulnerabilites and warnings to subtree of a …

dkcumming:
…target package

Passing `-t` followed by the name of a target package omits any vulnerabilities or warnings found that are not from packages in the subgraph formed from the target node. This is achieved by a depth first search for the vulnerable packages, from the target node.

Currently misses for the target name, and multiple targets returned for the target name are handled naively by ignoring flag and continuing with the full graph.

If `-t` is not provided, no change to complexity.
Providing `-t` increases complexity by $O(P \times (V' + E'))$ where $V'$ is the vertices in the subgraph (packages), $E'$ is the edges in the subgraph (dependencies), $P$ is the sum of all vulnerabilites and warnings found from the audit.


Shnatsel:
Thank you for the PR!

Could you clarify why such a feature is desirable? This introduces additional complexity to the codebase, and I'm not sure what are the use cases for it.

If we do decide to merge this, I'd also expect all TODOs to be done and tests to be added.

tarcieri:
We'd previously discussed this a bit over email, although I thought the intent was to enable it at the library-level, not necessarily something that would be in `cargo-audit`

rustsec/src/report.rs
```diff
@@ -39,15 +43,16 @@ pub struct Report {
 }
 
 impl Report {
-    /// Generate a report for the given advisory database and lockfile
-    pub fn generate(db: &Database, lockfile: &Lockfile, settings: &Settings) -> Self {
+    /// Generate a report for the given advisory database and lockfile, optional narrowing of subgraph from target package
+    pub fn generate(db: &Database, lockfile: &Lockfile, settings: &Settings, tree:&dependency::Tree, target:Option<&Package>) -> Self {
```
tarcieri, line 47: There are already two types for carrying additional state so you don't have to add parameters to this method: `Settings` and its `Settings::query` field which contains a `database::Query`.

I would think you could access the dependency tree via the `Lockfile`, and this feature could potentially set some option on the `database::Query`.

dkcumming:
Hi @Shnatsel, the motivation for the feature that auditing large packages can result in a large amount of vulnerabilities being output. Sometimes it is beneficial to isolate a particular part and see what vulnerabilities exist for that component of the package. Previously when auditing a large code base I was using `cargo audit` in combination with `cargo tree` to determine if the returned vulnerabilities were dependent in the subtree of a target package, but it felt appropriate to add the feature to `cargo audit` directly.

rustsec/src/report.rs
dkcumming: Yes, this makes more sense. I will do a refactor using `Query`

dkcumming:
Hi @Shnatsel @tarcieri, 
Sorry for the long time between commits. I have had other work. I implemented some solutions to the requested changes, however this push is to get some feedback from you. Some details of this push are:

Features:
- Refactored the parameters to make use of query.
- Added feedback to user if target package doesn't exist, or multiple packages match the identifier. I had a look at `cargo tree` for this as they had nice feedback for this. I pulled and edited some of the code (credited), extending the functionality useful for this case and stripping what wasn't relevant.

Known Problems:
- The location of the files is not in final place. As I was moving them I thought it better to get some feedback from you.
- I expect there is some sub optimal rust code in here, I am not an expert rust developer by any means, so feedback is appreciated.

cargo-audit/src/lockfile.rs
```diff
@@ -28,3 +29,141 @@ pub fn generate() -> rustsec::Result<()> {
...
 }
+
+// ! Package identifiers.
+// !
+// ! Adapted from Cargo's `package_id_spec.rs`:
+// !
+// ! <https://github.com/rust-lang/cargo/blob/master/src/cargo/core/package_id_spec.rs>
+// !
+// ! Copyright (c) 2014 The Rust Project Developers
+// ! Licensed under the same terms as the `cargo-lock` crate: Apache 2.0 + MIT
```
dkcumming, line 40: This does not belong here, but I was wondering what structure you would recommend.

dkcumming:
Further progression. I refactored and added test cases to show the restriction of output for different branches. These can be run in the from the test directory like so:
- **No filtering** 
    `multi_vulns$ ../../../../target/debug/cargo-audit audit -t multi_vuln` or 
    `multi_vulns$ ../../../../target/debug/cargo-audit audit`
-**Filter base64 branch**
    `multi_vulns$ ../../../../target/debug/cargo-audit audit -t base64`
-**Filter chrono branch**
    `multi_vulns$ ../../../../target/debug/cargo-audit audit -t chrono` or 
    `multi_vulns$ ../../../../target/debug/cargo-audit audit -t time` (for the time dependency vuln directly)

Previously I thought there was an issue with the layout when printing the tree "`The tree does not print from the target package node. I missed this as I only use the json implementation, but I will implement it.`". After using this a few times I realised the current output is correct as is.

@Shnatsel @tarcieri  
Are you able to provide feedback, or what you would need to see for a merge?
This change has been very useful for RV with audits, and we believe is likely useful to others who use `cargo audit`, it would be great to make it available.

dkcumming:
No problem, I see that some CI tests failed. Looks like related to the lockfile. Is that related to anything on my end that I need to address?

tarcieri:
You can try rebasing

dkcumming:
I synced the `main` branch with `upstream/main`. The `feature` branch was already merged with the `upstream/main`. Is this what you meant?

ACassimiro:
@Shnatsel @tarcieri I would just like to leave a ping here and say that this was a really helpful feature for going through specific packages in a huge repository. It saved me quite a bit of effort.

tarcieri:
I should have some time this weekend to take a look

dkcumming:
@tarcieri If there would be any benefit from a meeting to go over anything, I'm available pretty much anytime weekend or weekday for a meeting.

tarcieri:
@dkcumming can you rebase? that should fix the test failures

dkcumming:
@tarcieri  Should be updated now

tarcieri:
@dkcumming looks like there's a test failure

dkcumming:
@Shnatsel @tarcieri I haven't actually needed to use this feature on any projects since I have only been working with smaller trees since I did this. But I had some spare time this weekend to go and fix the problems that were on CI. If I get time another weekend I might add some test cases, but since I haven't needed this feature I am just going to put that on hold for now.

Shnatsel:
Thank you for updating the PR!

I admit I am hesitant to merge this, because we are struggling to maintain the code for the basic functionality as it is. Taking on more complex features would make it even harder, and is probably not something we should be doing at this stage, unless someone is willing to provide funding to the project. @tarcieri WDYT?

PR #780, opened 2023-01-04
Support sparse registry references in Lockfiles

rbtcollins:
Closes #604

Signed-off-by: Robert Collins <robert.collins@cognite.com>

PR #870, opened 2023-04-24
cargo-lock v9.0.0

tarcieri:
### Added
- Implement `From<Name>` for `String` ([#776])
- Support sparse registry references in `Lockfile`s ([#780])

### Changed
- Mark `SourceKind` as `#[non_exhaustive]` ([#793])
- Use `Display` for `io::ErrorKind`; MSRV 1.60 ([#794])
- Bump `toml` to 0.7 ([#800], [#805])
- Improvements to the `cargo lock tree` subcommand ([#860])

### Fixed
- `Source::is_default_registry` for sparse index ([#859])

[#776]: https://github.com/RustSec/rustsec/pull/776
[#780]: https://github.com/RustSec/rustsec/pull/780
[#793]: https://github.com/RustSec/rustsec/pull/793
[#794]: https://github.com/RustSec/rustsec/pull/794
[#800]: https://github.com/RustSec/rustsec/pull/800
[#805]: https://github.com/RustSec/rustsec/pull/805
[#859]: https://github.com/RustSec/rustsec/pull/859
[#860]: https://github.com/RustSec/rustsec/pull/860

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,7 +185,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "57ba7d807ad2850c436750154dbc442eb8611527deb6dfcc4e5e0bd10d5fbf28"
 dependencies = [
- "cargo-lock 8.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cargo-lock 8.0.3",
  "semver",
  "serde",
  "serde_json",
@@ -281,7 +281,7 @@ dependencies = [
  "auditable-info",
  "auditable-serde",
  "binfarce",
- "cargo-lock 8.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cargo-lock 8.0.3",
  "clap",
  "home",
  "once_cell",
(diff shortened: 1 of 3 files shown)


tarcieri:
@Shnatsel it looks like you removed the overrides for `cargo-lock` in 5ea5b843.

This kind of defeats the point of having a monorepo. `cargo-lock` has diverged in API from what the other crates are using and there are some API changes needed to support the new version.

tarcieri:
Oh now I see, `auditable-serde` lives out-of-tree, but depends on `cargo-lock`.

@Shnatsel that makes for a somewhat nasty circular dependency. Any chance you can pull `auditable-serde` into this repo?

Shnatsel:
I suppose I could move the impl to the `auditable-serde` crate? IIRC it should be allowed under the orphan rules. Would that resolve the issue?

tarcieri:
The impl is already there.

The problem is auditable-serde depends on cargo-lock, and cargo-audit depends on auditable-serde, which is a circular dependency across the `rustsec` and `cargo-auditable` repos

Shnatsel:
I've given you publishing rights to the entire (?) cargo-auditable dependency tree so you could publish updates if needed.

tarcieri:
It would technically work, yes, but that feels like the wrong dependency relationship. It also makes `auditable-serde` part of `cargo-lock`’s public API

Shnatsel:
Any other ideas, aside of merging the repos?

tarcieri:
The only other one is splitting `cargo-lock` (back) out into its own repo.

However, avoiding these kinds of problems was the motivation for making a monorepo in the first place.

PR #875, opened 2023-04-27
Bump `cargo-lock` to v0.9 + auditable deps

tarcieri:
Also upgrades the following `auditable` dependencies which transitively depend on `cargo-lock` v0.9:

- `auditable-info` v0.7
- `auditable-serde` v0.6


Cargo.toml
```diff
@@ -8,3 +8,10 @@ members = [
     "quitters",
     "rustsec"
 ]
+
+[patch.crates-io]
+cargo-lock = { path = "./cargo-lock" }
+cvss = { path = "./cvss" }
+platforms = { path = "./platforms" }
+quitters = { path = "./quitters" }
+rustsec = { path = "./rustsec" }
```
tarcieri, line 17: This should be able to handle circular dependencies a bit better

Issue #832, opened 2023-03-20
CI: cargo audit tests take a long time


pinkforest:
test auditable_binary_with_vulnerabilities_fails has been running for over 60 seconds
test auditable_binary_without_vulnerabilities_passes has been running for over 60 second

~13+ minutes on all platforms:

Windows: https://github.com/rustsec/rustsec/actions/runs/4467658995/jobs/7847399841?pr=831#step:6:282
MacOS: https://github.com/rustsec/rustsec/actions/runs/4467658995/jobs/7847397875#step:6:279

Some GitHub red herring or something we may need to cache somewhere perhaps ?

EDIT: Doesn't seem to be red herring - 2nd run doing the same on all platforms each over 10 mins:

https://github.com/rustsec/rustsec/actions/runs/4467898232/jobs/7847945238?pr=831

Shnatsel:
Just using the default DB location should help, actually.

Shnatsel:
Actually it's much simpler than that: the caching seems to be broken.

https://github.com/rustsec/rustsec/actions/runs/4485191049/jobs/7886488822?pr=835#step:12:2

Shnatsel:
#839 shaves 5 minutes off the CI time, but caching is still broken.

tarcieri:
Might be worth updating #551 and see if the issues have resolved themselves in the interim

Shnatsel:
We've gotten CI times under control. They are 1-2 minutes now. Closing.

PR #883, opened 2023-05-09
Update integration test to lower runtimes

simonsan:
Does this have any downsides? It should speed up the integration tests overall?

could also help with #832

related: https://github.com/iqlusioninc/abscissa/issues/845

### DIFF
--- a/cargo-audit/tests/acceptance.rs
+++ b/cargo-audit/tests/acceptance.rs
@@ -26,7 +26,7 @@ static ADVISORY_DB_DIR: Lazy<TempDir> = Lazy::new(|| TempDir::new().unwrap());
 /// be multithreaded invocations as `cargo test` executes tests in
 /// parallel by default.
 pub static RUNNER: Lazy<CmdRunner> = Lazy::new(|| {
-    let mut runner = CmdRunner::new(env!("CARGO_BIN_EXE_CARGO_AUDIT"));
+    let mut runner = CmdRunner::new(env!("CARGO_BIN_EXE_cargo-audit"));
     runner.arg("audit").arg("--db").arg(ADVISORY_DB_DIR.path());
     runner.exclusive().capture_stdout().capture_stderr();
     runner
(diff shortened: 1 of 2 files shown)


cargo-audit/tests/acceptance.rs
```diff
@@ -26,9 +26,9 @@ static ADVISORY_DB_DIR: Lazy<TempDir> = Lazy::new(|| TempDir::new().unwrap());
 /// be multithreaded invocations as `cargo test` executes tests in
 /// parallel by default.
 pub static RUNNER: Lazy<CmdRunner> = Lazy::new(|| {
-    let mut runner = CmdRunner::default();
+    let mut runner = CmdRunner::new(env!("CARGO_BIN_EXE_CARGO_AUDIT"));
```
tarcieri, line 29: What is this environment variable?

Shnatsel: It seems this should be using `std::env::var_os` instead of `env!`. 

I doubt this is going to be any faster, but I'd be happy to be proven wrong.

Shnatsel: Panic - it's supposed to be always set: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

Shnatsel: Nevermind, I got it all wrong. It should be `env!(CARGO_BIN_EXE_cargo-audit)`

Shnatsel:
**TL;DR:** this is a worthwhile change, but I'd like to see it on the framework level (i.e. in `abscissa`) rather than fix up every user individually.

I've compared this against the main branch, measuring the second run of the tests (once the crates.io index cloning is done, which takes up the vast majority of the time), and it takes 50 seconds regardless of these changes.

With an up-to-date crates.io index, running `cargo clean; cargo test --release --all-features --no-run` and then measuring `cargo test --release --all-features` shows the new approach to be faster, because `cargo run` is invoked without `--release` and genuinely needs compile the debug build instead of using the release build that's already been compiled.

Contrary to @simonsan's suspicion, the compilation is performed exactly once, and cached after that. There's no recurring compilation assuming the Cargo cache works properly.

tarcieri:
>  I had several cmd mutex poisened errors when running tests with the default configuration of CmdRunner for rustic in the beginning.

This will occur if a test fails while the mutex is held.

It should probably use `panic::catch_unwind` to prevent that.

PR #914, opened 2023-07-08
Case-insensitive search on website

### LINKED ISSUE #913: Make website search field case-insensitive
Searching for `Foobar` shows no results even if there are advisories for the crate `foobar`. I see no reason for this search to be case-sensitive?

This is a more noticeable problem on mobile, where words are auto-capitalized. It lead to me not finding any advisory on crate that had two (https://github.com/Gilnaa/memoffset/issues/76).

amousset:
Fixes #913


Shnatsel:
Thanks!

PR #922, opened 2023-08-02
change `git2` to `gix` and remove `openssl` dependency

### LINKED ISSUE #750: Get rid of OpenSSL

Byron:
This PR is used to align with the maintainers on how to proceed with this task.

Switching to `gix` in all dependencies is now possible as `crates-index` v2.0 uses `gix`.
This in turn allows to use alternative backends that don't use `openssl`.

### Tasks

* [x] upgrade `admin`  and `rustsec` to `crates-index` v2.0 with equivalent configuration
* [ ] switch `crates-index` to using the `reqwest` backend for HTTPS
* [ ] convert `git2` code to `gix`
* [ ] see if `git-performance` works for `rustsec` (depends on required platform support)

### Caveats

Currently, when using the pure-Rust https backend, `http.proxy` isn't supported. This could be implemetented
easily if you need it (or other features related to git's `http` configuration).

### Issues

* fixes #750
* should reoslve #466

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2654,7 +2654,7 @@ version = "0.27.0"
 dependencies = [
  "cargo-edit-9",
  "cargo-lock",
- "crates-index 0.19.13",
+ "crates-index 2.0.0",
  "cvss",
  "fs-err",
  "git2",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,6 +57,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+dependencies = [
+ "cfg-if",
+ "getrandom",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -91,6 +103,12 @@ name = "arc-swap"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
 name = "askama"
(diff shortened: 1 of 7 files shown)


Byron:
This is a draft that I'd like to use to align with the maintainers. Is my task-list what you had in mind? Are the caveats an issue for you? Is this something you would want me to do even? 

I am looking forward to hearing your thoughts :).

Cargo.lock
```diff
@@ -585,6 +688,36 @@ dependencies = [
...
+checksum = "961ba061c9ef2fe34bbd12b807152d96f0badd2bebe7b90ce6c8c8b7572a0986"
+dependencies = [
+ "cc",
+ "libc",
+ "libz-sys",
+ "openssl-sys",
+ "pkg-config",
+ "vcpkg",
+ "winapi",
+]
```
tarcieri, line 719: Are there any options for Rust HTTP clients?

Byron: I have added a commit that removes `curl` in favor of the only other available HTTPS backend: `reqwest`.

Shnatsel: We should switch from `crates-index` to `tame-index` as covered [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/146229-wg-secure-code/topic/removing.20OpenSSL.20from.20.60cargo.20audit.60), following the same switch in `cargo-deny`.

Byron: Ah, great! Just today I learned about `tame-index` independently of that and it looks like it's the way to go!

I am happy to help with/implement the switch from `git2` to `gix` if you want me to, please feel free to ping me in GitHub or on zulip.

Shnatsel:
Thank you very much for the PR!

There is some Git tree-walking code in `admin` used for OSV export, I don't think it will be necessary to convert immediately since it is only used on CI. It should not be blocking shipping of `rustls` and `gix` builds to users.

I would like to understand more about the 2.0 version of `crates-index`. Last time I checked, in the git main there was an option to use sparse index, but the clients had to bring their own HTTP client. `tame-index` bundles one, so it could be more convenient for us (at the expense of pulling in the whole `reqwest` dependency tree, but we do need HTTP/2 support so `ureq` et all will not cut it). Does `crates-index` 2.0 expose a higher level API that doesn't require bringing your own HTTP client? If so there is probably no need for using `tame-index`.

Byron:
> Thank you very much for the PR!

I would have been my pleasure, but I will try again by helping with the `git2` to `gix` conversion :).

> I would like to understand more about the 2.0 version of crates-index. Last time I checked, in the git main there was an option to use sparse index, but the clients had to bring their own HTTP client. tame-index bundles one, so it could be more convenient for us (at the expense of pulling in the whole reqwest dependency tree, but we do need HTTP/2 support so ureq et all will not cut it). Does crates-index 2.0 expose a higher level API that doesn't require bringing your own HTTP client? If so there is probably no need for using tame-index.

I think `crates-index` v2.0 wants to be `tame-index`, which has seen more development time and improvements as well. I dare to say that it's tailored to be used with `gix`, it even packages the `reqwest` client to support HTTPS because `gix` does the same (if configured).

For now I am done with `crates-index` as I have it a point where it's good for my single use-case, but I would definitely be seeing myself in the boat of switching it over to `tame-index` for ease-of-use. Particularly because sparse-indices are easy to use, I wanted that, too.

In short, I think that means there is no need to look at `crates-index`, I think `tame-index` is the safer bet and more advanced as well.

CC @Jake-Shadle due to the praise for `tame-index` :)

Issue #868, opened 2023-04-24
cargo-audit fails on sparse registries


waltronix:
`cargo-audit` fails to parse `Cargo.lock` files with when package source starts with `sparse+`.

Internally, we are using a sparse private registry as proxy for crates-io.

When trying to run `cargo-audit` I get the following error.

```
error: not found: Couldn't load Cargo.lock: I/O operation failed: parse error: parse error: unsupported source protocol: sparse for key `package.source` at line 366 column 1
```

I could reproduce the issue with a minimal sample where I use the following `config.toml`

```toml
[registries.sparse_sample]
index = "sparse+https://index.crates.io/"
```

with a `Cargo.toml`

```toml
[package]
name = "sample"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
log = {version = "0.1", registry = "sparse_sample"}
```

In `Cargo.lock`, the package definitions look like:

```toml
[[package]]
name = "log"
version = "0.1.10"
source = "sparse+https://index.crates.io/"
checksum = "527343c5c05a389bbd6818d495e5d7577feb4dd51ca698c158edabf398a3a4bc"
dependencies = [
 "regex",
]
```

The issue already seems to be fixed in `cargo-lock`, but `cargo-audit` does not already use the fixed version here.

tarcieri:
I opened https://github.com/rustsec/rustsec/pull/870 but I'm in dependency hell

Shnatsel:
Just needs a changelog and the actual release, I'll try to get it done later today. Thanks for the reminder.

tarcieri:
Upgrading `cargo-lock` in `rustsec` is a SemVer breaking change.

That needs a minor version bump at the very least.

Shnatsel:
I believe it needs a major version bump, because `cargo_lock::Lockfile` is [re-exported as-is](https://docs.rs/rustsec/0.26.5/rustsec/struct.Lockfile.html). I've included a major bump in the release PR, see #885

tarcieri:
That's a minor version bump, but yeah that's what I had in mind

mikhasd:
Thanks for the new release. I can now scan my internal project.

Note that the tool prints a warning though:
> `warning: couldn't open crates.io index: registry: 'sparse+https://INTERNAL/index/' is not a valid registry url`

mikhasd:
> Is that coming out of `crates-index`?

[...]

Issue #866, opened 2023-04-21
Respect `[registries.crates-io] protocol = "sparse"` set in cargo config


Logarithmus:
I have this in my global cargo config:
```toml
[registries.crates-io]
protocol = "sparse"
```

Would like `cargo audit` to respect it instead of forcibly downloading the slow git index. From the code I can see you already support sparse registries, but only if they are specified in `Cargo.lock`. But as you can see [here](https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html#:~:text=Cargo.lock%20files%20will%20continue%20to%20reference%20the%20existing%20crates.io%20index%20on%20GitHub%20to%20avoid%20churn..), "Cargo.lock files will continue to reference the existing crates.io index on GitHub to avoid churn". So your current approach doesn't work.

tarcieri:
We use the `crates-index` crate to fetch the index: https://github.com/frewsxcv/rust-crates-index

Please see the tracking issue there: https://github.com/frewsxcv/rust-crates-index/issues/87

Logarithmus:
@tarcieri they added sparse registry support, but only for local cache, it can't fetch anything from the network. So I guess this is still blocked.

Issue #846, opened 2023-03-25
advisory-db: CI linter Takes 1m 30 sec each PR


pinkforest:
Every PR in advisory-db the linter takes 1m 30s each time.

https://github.com/rustsec/advisory-db/actions/runs/4519170146/jobs/7959502197#step:5:1

I had a mild thought (I should have known better) that it might have been the 500+ log lines it spits out but that's not the case.

The runner just runs it slowly.

Would it be in PR's feasible to reduce the linter to the files that were touched as part of the PR ?

On push it could still stay as all files - or even have a periodic run sans every push instead.

tarcieri:
It's slow because it's communicating with the crates.io API.

That said, I don't think 1m30s is a terribly long time?

pinkforest:
ah ofcourse - so that's the crates-index which transparently pulls the index - didn't we have this cached now ?

e.g. didn't we fix this caching issue in:
https://github.com/rustsec/rustsec/issues/832

And could use the same cachign in advisory-db

tarcieri:
It's communicating directly with the crates.io API.

Perhaps that communication can be replaced with `crates-index`.

pinkforest:
The linter doesn't seem to communicate with the API - it's already using the database via crates-index ?
https://github.com/rustsec/rustsec/blob/main/admin/src/linter.rs#L155

tarcieri:
Aah, never mind, guess that already happened.

Then I'm not sure why it's so slow.

pinkforest:
No caching. Should be easy to copy caching from here to there me thinks.

pinkforest:
Looks like we need to use [Index::with_path](https://docs.rs/crates-index/latest/crates_index/struct.Index.html#method.with_path) and put the index repo into cache - I'll see if crates-index could get rid of git2 dependency in case of local path.

Shnatsel:
I think putting the index into the CI cache should be sufficient. Incremental updates to it should be pretty quick.

Regarding getting rid of `git2`, the upstream is interested in migrating to `gitoxide` and @BlackHoleFox is working on a PR.

pinkforest:
Yeah true simple incremental works - the crates.io-index gets re-rolled every three months due to sheer size of the history and needs to handle the pull in a way that it doesn't break it.

Issue #790, opened 2023-01-12
Fetching the crates.io index is slow


fenollp:
I am seeing it being stuck on this last line:
```
λ cargo audit --ignore RUSTSEC-2020-0071
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 480 security advisories (from /home/pete/.cargo/advisory-db)
    Updating crates.io index
```

tarcieri:
That seems like an issue with the `crates-index` crate

Shnatsel:
This is an issue that Cargo itself also faces. The root if it is in `libgit2`'s inefficient delta resolution algorithm, combined with the large size of crates.io repository.

This will be fixed by migrating to `gitoxide` which is planned as part of #750

PR #923, opened 2023-08-03
Transition from `crates-index` -> `tame-index`

Jake-Shadle:
Resolves: #846 - If the crates.io sparse index is used (ie user is using cargo 1.70.0+ or 1.68.0+ with `CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse` then the git clone/fetch will be eliminated and only requested crates will be fetched from the HTTP index, drastically improving index 'fetch'
Resolves: #790 - Ditto
Resolves: #866 - `tame-index` respects the user's [environment](https://github.com/EmbarkStudios/tame-index/blob/d12585d90c4a9df26c2afe0c1678c77b99b39548/src/index/location.rs#L48-L103)
Resolves: #868 - Ditto
Resolves: #292 - ...I think? The index fetching at least attempts to use [all the same authentication](https://github.com/EmbarkStudios/tame-index/blob/d12585d90c4a9df26c2afe0c1678c77b99b39548/src/index/git_remote.rs#L54-L63) as git does
Part of: #750 

I'll make a follow up PR to this that replaces `git2` with `gix` which should finish #750, among others.

I removed registry.rs since it was completely unused at least in this repo, but I can add it back if that was overstepping.

Unfortunately I don't believe there are any tests around the index support, and the current way that the code is structured makes testing harder than necessary, but let me know if you want me to add some tests. One of the design goals of `tame-index` that differentiates it from `crates-index` is that the API was explicitly designed to be able to easily be used in testing, eg. here is a [cargo-deny test](https://github.com/EmbarkStudios/cargo-deny/blob/16d0db8a448d9299b9c52884034227a603f5abc1/tests/advisories.rs#L450C1-L619C2) that just ensures that it can both source replacement of crates.io as well as yank checking work correctly that can still be run in parallel with any other tests since `tame-index` is made to default to cargo's defaults, but allows easy overriding of the local disk locations if desired to segment test data/snapshots.


Jake-Shadle:
Oh, this also bumps MSRV to 1.67.0, which is the minimum supported by gix/tame-index

Shnatsel:
I've noticed that when using the sparse index, the scanning for yanked crates is very slow. On this very branch it takes 55 seconds on my machine for ~400 dependencies, and the time does not go down with subsequent runs. With git protocol (also on this branch) it only takes 2 seconds if the index is already cloned.

The stack chart shows the thread mostly waiting on something, presumably the network responses: https://share.firefox.dev/47dRpAH

It appears that requests for crates.io data are made sequentially, so the execution time is dominated by the latency from my machine to crates.io multiplied by the number of packages (~400 in this repo). What Cargo does instead is request all 400 packages up front, which uses HTTP/2 multiplexing and resolves in only a few seconds. Would it be possible to use this approach here as well?

Jake-Shadle:
Yes, that would be trivial since the sparse index can be used in parallel, though I'd need to make modifications due to how the current code works.

rustsec/src/cached_index.rs
```diff
@@ -6,6 +6,28 @@ use crate::{
...
+
+enum Index {
+    Git(tame_index::index::RemoteGitIndex),
+    SparseCached(tame_index::index::SparseIndex),
+    SparseRemote(tame_index::index::RemoteSparseIndex),
+}
+
+impl Index {
+    #[inline]
+    fn krate(&self, name: &package::Name) -> Result<Option<tame_index::IndexKrate>, Error> {
```
Shnatsel, line 19: To make sparse index fetch not terribly slow, this function will have to be removed. Instead need an API that accepts a bunch of package names all at once, and returns a map of package names to package metadata.

Jake-Shadle:
Ahh, ok

Shnatsel:
Excellent, the time is way better now! I'm seeing 1.8s for the scan instead of 55s taken previously.

Shnatsel:
I've [taken my own stab](https://github.com/Shnatsel/rustsec/commit/729eafa3e23702c9939acfcf336e7f9a1272cf70) at async-ifying this, but this didn't help, and a [supposed solution](https://github.com/Shnatsel/rustsec/commit/7a9bf763f1fe0e1b3b9efc54f38c7514a74cac2e) just landed me in lifetime hell :sweat_smile: 

I'm glad that you've cracked it. `tame-index` is a much better place for this kind of function anyway.

Jake-Shadle:
Yah, it's been a while since I've done async and was happy to find out that tokio now has a [much nicer](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html) way to spawn multiple independent tasks that made the implementation much simpler.

rustsec/src/cached_index.rs
```diff
@@ -36,47 +85,134 @@ impl CachedIndex {
...
+                            .into_iter()
+                            .map(|p| p.as_str().to_owned())
+                            .collect(),
+                        true,
+                        REQUEST_TIMEOUT,
+                    )
+                    .map_err(|err| {
+                        format_err!(
+                            ErrorKind::Registry,
+                            "unable to acquire tokio runtime: {}",
```
Shnatsel, line 155: I believe this error message about Tokio runtime is misleading.

Shnatsel:
I'm about to do a more thorough review pass, but one thing stood out to me before the switch from `rayon` to `tokio`: the data from the sparse index was always fetched from the network, never read from the local disk cache. The code seems to imply that the disk cache is supposed to be used. Is that a bug?

rustsec/src/cached_index.rs
```diff
@@ -36,47 +85,134 @@ impl CachedIndex {
     }
 
     /// Open the local crates.io index
+    ///
+    /// If this opens a git index, it allows reading of index entries from the
+    /// repository
+    ///
+    /// If this is a sparse index, it only allows reading of index entries that
+    /// are already cached locally
```
Shnatsel, line 93: This documentation confuses me. What happens if the crate I open is not in the cache?

Ideally there should be an API that transparently uses the local on-disk cache and fetches remote data only when necessary, where "when necessary" is determined in the same way as in Cargo itself.

rustsec/Cargo.toml
```diff
@@ -24,8 +24,10 @@ toml = "0.7"
 url = { version = "2", features = ["serde"] }
 
 # optional dependencies
-cargo-edit = { version = "0.9", package = "cargo-edit-9", optional = true, default-features = false,  features = ["upgrade"] }
-crates-index = { version = "0.19", optional = true }
+cargo-edit = { version = "0.9", package = "cargo-edit-9", optional = true, default-features = false, features = [
+    "upgrade",
+] }
+tame-index = { version = "0.3", features = ["git", "sparse"], optional = true }
```
Shnatsel, line 30: I just wanted to say I appreciate `git` and `sparse` being separate features that can be toggled individually :heart:

cargo-audit/src/auditor.rs
```diff
@@ -243,6 +243,21 @@ impl Auditor {
...
+                    .iter()
+                    .filter_map(|pkg| {
+                        pkg.source
+                            .as_ref()
+                            .filter(|s| s.is_default_registry())
+                            .map(|_s| &pkg.name)
+                    })
+                    .collect(),
+            ) {
+                status_err!("failed to execute populate_cache, index metadata may be missing or stale when checking for yanked packages: {}", err);
```
Shnatsel, line 258:
```suggestion
                status_err!("Failed to download crates.io index: {}\nData may be missing or stale when checking for yanked packages.", err);
```

rustsec/src/cached_index.rs
Jake-Shadle: If a crate is not in the index cache it will return `None`.

Cargo will always do a remote operation unless in offline mode, as the only way to check if the local cache entry is up to date is to send the etag (or last-modified but AFAICT cargo only uses the e-tag) of the cache entry to either get an updated entry or a `NOT_MODIFIED`. The comment is just saying that it works the exact same as the git index in the sense that only local information will be used, the difference is that in the git case crates that haven't been previously queried by cargo/other tools _might_ still be present in the git blobs, and thus readable (if outdated), whereas in the sparse case they won't exist in the local cache.

Shnatsel: Got it. Thank you!

rustsec/src/cached_index.rs
Shnatsel: This is handling an error returned by the function `krates_blocking`, right? For example, a timeout may cause it? Or am I misunderstanding the code?

Jake-Shadle: How so? In this case the error should never occur since the runtime is being created just above this function, but the error is only possible the in the case that https://docs.rs/tokio/1.29.1/tokio/runtime/struct.Handle.html#method.try_current fails.

Jake-Shadle: All timeouts or other errors would be particular to each individual crate with how the current code works.

Shnatsel: Ah! I see. I misunderstood the code then. Thank you for clarifying!

Shnatsel (approved):

Shnatsel:
Thank you very much! I'm really excited to merge this!

Issue #555, opened 2022-05-06
Failed to fetch private git repo


danielhaap83:
Hi,
fetching a private git repository fails using `https` and gits `credential.helper = store`.

Example code:
```rust
fn main() {
    let tmp_dir = tempdir().unwrap();
    let url = "https://github.com/danielhaap83/private-repo.git";

    Repository::fetch(url, tmp_dir.path(), true).unwrap();
}
```

**Error**
```sh
cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/crash_rustsec`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Repo, msg: "failed to authenticate when downloading repository\nattempted to find username/password via git's `credential.helper` support, but failed" }', src/main.rs:8:50
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Expected**
Clones my private repository at tmp directory.

**Authentication Verified**
Git clone using executable from shell works.
```sh
cd /tmp
/tmp 🐠  git clone https://github.com/danielhaap83/private-repo.git
Cloning into 'private-repo'...
remote: Enumerating objects: 6, done.
remote: Counting objects: 100% (6/6), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 6 (delta 0), reused 0 (delta 0), pack-reused 0
Unpacking objects: 100% (6/6), 1.20 KiB | 245.00 KiB/s, done.
```

**Prerequisite**
1. repository with visibility set to `private`
2. git credential helper store configured: `git config --global credential.helper store`
3. git credential helper store enabled
- clone the configured repository via git to create the store `git clone https://github.com/xxx/private-repo`
- enter username and password. I used my private access token as password

**Versions:**
 - `Ubuntu 20.04.4 LTS`
 - `git executable version 2.25.1`
 - `rustsec 0.25`

Issue #494, opened 2021-11-26
cargo install cargo-audit ailes with invalid version 0 on git_proxy_options; class=Invalid (3)


sandeep-gh:
I am on OpenBSD and had to update the system (git, cmake, etc) to install rust. However, after installation cargo update is failing: 
```
pi4bsdm2% cargo install cargo-audit --features vendored-libgit2
    Updating crates.io index
error: failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  invalid version 0 on git_proxy_options; class=Invalid (3)
```
The issue is related to https://github.com/rustsec/rustsec/issues/431 but for me the fix is causing the same problem. 

The libgit2 library:
```
pi4bsdm2% pkg_info libgit2
Information for inst:libgit2-1.3.0

Comment:
the Git library, take 2

Required by:
rust-1.56.1

WWW: https://libgit2.org/
```

tarcieri:
This looks like an error with `cargo` and appears to be unrelated to `cargo-audit` itself

klemensn:
This is likely caused by OpenBSD's recent `libgit2` update to `1.3.0`.

I hit this very same error with `cargo update` in another repository:
```
$ cargo update
    Updating crates.io index
error: failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  invalid version 0 on git_proxy_options; class=Invalid (3)
```

Forcing older `libgit2-1.1.0` fixed it:
```
$ LD_PRELOAD=/usr/local/lib/libgit2.so.13.0 cargo update
    Updating crates.io index
    [ update succeeds ]
```

Having done that once, i.e. with a fresh index on disk, the newer `libgit2` also works again...
```
$ ldd `which cargo` | grep git
        000000375faf8000 000000375fcc8000 rlib  0    1   0      /usr/local/lib/libgit2.so.14.0
$ cargo update
cargo update -p cpal
    Updating git repository `...`
    Updating crates.io index
```

No idea why.
Perhaps some state had to be fetched and triggered the error with newest `libgit2`?
Now that the index is up-to-date this particular code path isn't hit anymore and that is why both versions now succeed?

Moving away the index seems to confirm this theory:
```
$ mv .cargo/registry/index/github.com-1ecc6299db9ec823{,.old}
```
```
$ cargo update
    Updating git repository `...`
    Updating crates.io index
error: failed to get `...` as a dependency of package `...`

Caused by:
[...]

Issue #490, opened 2021-11-17
Fetching rustsec DB does not handle repo contention


epage:
Similar to the crate registry (see #489), it looks like fetching the rustsec DB is vulnerable to failing on the same race condition.  We should support retries, whether inside rustsec or by providing enough context to the user so they know they should attempt a retry (looks like all we can tell is there is a "repo" error).

Shnatsel:
Reopening since the approach from #925 runs into issues with stale lockfiles being left if the process is interrupted, causing all subsequent invocations to fail. It's possible to paper over this with signal handlers to some extent but currently the `gix` signal handlers are [not sophisticated enough](https://github.com/rustsec/rustsec/pull/934#discussion_r1288376326).

Shnatsel:
Upon further investigation we found the approach in #925 to be sufficient. I've opened https://github.com/rustsec/rustsec/pull/934 to make it more clear what's happening in case a stale lockfile is present, but the issue described here is fixed.

PR #925, opened 2023-08-08
Transition from git2 -> gix

Jake-Shadle:
This PR replaces git2 with gix, in rustsec, which should fix a bunch of issues. Note that there is still a dependency on git2/openssl via cargo-edit, but that's out of scope for this PR.

I tried to do a straight replacement, and all the tests pass, but I don't know how comprehensive the tests are in this repo. I say that as a caveat that I've spent more time on this than I had originally intended, so if there is going to need to be a bunch of feedback fixup or the like it would be better if someone took this PR over since I'm moving on to the other stuff and won't have time to address a lot of feedback.

I'll add comments to this PR to help review.

Resolves: #555 (gix should automatically do the correct thing)
Resolves: #494 (libgit2 no longer an issue, at least for cargo-audit itself, could still be a problem for cargo-edit)
Resolves: #490 (an fs lock now protects the opening of an advisory repo)
Resolves: #466 (git2 no longer used)


rustsec/Cargo.toml
```diff
@@ -28,10 +28,18 @@ cargo-edit = { version = "0.9", package = "cargo-edit-9", optional = true, defau
     "upgrade",
 ] }
 tame-index = { version = "0.3", features = ["git", "sparse"], optional = true }
-git2 = { version = "0.16.1", optional = true }
 home = { version = "0.5", optional = true }
-humantime = { version = "2", optional = true }
-humantime-serde = { version = "1", optional = true }
+time = { version = "0.3", default-features = false, features = ["formatting", "serde"], optional = true }
```
Jake-Shadle, line 32: `time` is already a dependency due to `gix`, and has the formatting stuff for serde already so no need for the humantime stuff.

rustsec/src/error.rs
```diff
@@ -131,14 +131,6 @@ impl From<fmt::Error> for Error {
...
 
-#[cfg(feature = "git")]
-#[cfg_attr(docsrs, doc(cfg(feature = "git")))]
-impl From<git2::Error> for Error {
-    fn from(other: git2::Error) -> Self {
-        format_err!(ErrorKind::Repo, &other)
-    }
-}
-
 impl From<io::Error> for Error {
```
Jake-Shadle, line 134: So unfortunately `gix` doesn't have a singular Error type, but rather...hundreds of unique errors. This is a bit tedious, but does mean that all the places that throw errors now have a human (me) supplied contextual string for them rather than that info getting lost in the previous way of just straight converting between error types.

rustsec/src/repository/git/commit.rs
```diff
@@ -40,40 +37,64 @@ pub struct Commit {
...
+
+        if summary.is_empty() {
+            return Err(format_err!(
+                ErrorKind::Repo,
+                "no commit summary for {}",
+                commit_id
+            ));
+        }
+
+        let (signature, signed_data) = if let Some(sig) = cref.extra_headers().pgp_signature() {
```
Jake-Shadle, line 71: As noted, I opened https://github.com/Byron/gitoxide/pull/971 to add this to gix itself, but in the meantime this is just a more inefficient way to get the signature/signed data. This could also just be removed since it's not currently validated at all.

rustsec/src/repository/git/modification_time.rs
```diff
@@ -31,69 +31,163 @@ impl GitModificationTimes {
...
-            humantime::format_rfc3339(std::time::UNIX_EPOCH.add(duration_from_epoch)).to_string();
-        time.truncate(10);
-        time
+    fn gix_time_to_date(timestamp: &Time) -> Date {
+        let odt = crate::repository::git::gix_time_to_time(*timestamp);
+        let date = odt.date();
+
+        format!(
+            "{:0>4}-{:0>2}-{:0>2}",
+            date.year(),
```
Jake-Shadle, line 186: The time formatting stuff is annoying to use, this is just a manual version since the date part is trivial.

rustsec/src/repository/git/repository.rs
```diff
@@ -72,64 +69,99 @@ impl Repository {
...
+
+        let (mut repo, fetch_outcome) = open_or_clone_repo()?;
+
+        if let Some(fetch_outcome) = fetch_outcome {
+            tame_index::utils::git::write_fetch_head(
+                &repo,
+                &fetch_outcome,
+                &repo.find_remote("origin").unwrap(),
+            )?;
+        } else {
```
Jake-Shadle, line 154: Note that in both cases we write a `FETCH_HEAD` (gix doesn't yet do this). IMO this should be the timestamp used for staleness detection, but since I wanted to do a straight port it still uses the commit time of the HEAD commit.

rustsec/src/repository/git/repository.rs
```diff
@@ -163,6 +200,116 @@ impl Repository {
...
+                format_err!(ErrorKind::Repo, "failed to set `committer.name`: {}", err)
+            })?;
+        // Note we _have_ to set the email as well, but luckily gix does not actually
+        // validate if it's a proper email or not :)
+        config
+            .set_raw_value("committer", None, "email", "")
+            .map_err(|err| {
+                format_err!(ErrorKind::Repo, "failed to set `committer.email`: {}", err)
+            })?;
+
```
Jake-Shadle, line 237: Particularly in CI scenarios, this is needed as otherwise gix will fallback to the global config, or fail if there is not a committer set, when doing fetch and/or reflog updates.

Shnatsel:
cc @Byron to see what a real-world conversion from `libgit2` to `gix` looks like.

quitters/src/lib.rs
```diff
@@ -68,6 +68,7 @@ pub fn versions(data: &[u8]) -> BTreeSet<(&str, Version)> {
     // Basically streaming the data. This requires special handling of the start and end,
     // so either needs a state-machine-based parser like nom or capping the possible match length.
     // The latter is doable but only makes sense if it turns out that the current approach is too slow.
+    #[allow(clippy::invalid_regex)]
```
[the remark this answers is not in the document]
Shnatsel: This is a false positive in Clippy. Or rather, it calls out a valid issue for when the regex is used in unicode mode, but we disable unicode support so this is not an issue for us.

We do run Clippy on CI so this opt-out is necessary.

rustsec/src/repository/git/modification_time.rs
```diff
@@ -31,69 +31,163 @@ impl GitModificationTimes {
...
         let mut mtimes: HashMap<PathBuf, Time> = HashMap::new();
         let mut ctimes: HashMap<PathBuf, Time> = HashMap::new();
-        let repo = git2::Repository::open(repo.path())?;
-        let mut revwalk = repo.revwalk()?;
-        revwalk.set_sorting(git2::Sort::TIME)?;
-        revwalk.push_head()?;
-        for commit_id in revwalk {
-            let commit_id = commit_id?;
-            let commit = repo.find_commit(commit_id)?;
+
```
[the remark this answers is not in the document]
Shnatsel: No divergences with the previous `git2` implementation :+1:

rustsec/src/osv/advisory.rs
```diff
@@ -248,10 +247,10 @@ fn timeline_for_advisory(versions: &crate::advisory::Versions) -> OsvJsonRange {
     }
 }
 
-fn git2_time_to_rfc3339(git_timestamp: &git2::Time) -> String {
-    let unix_timestamp: u64 = git_timestamp.seconds().try_into().unwrap();
-    let duration_from_epoch = std::time::Duration::from_secs(unix_timestamp);
-    humantime::format_rfc3339(std::time::UNIX_EPOCH.add(duration_from_epoch)).to_string()
+fn git_time_to_rfc3339(time: gix::date::Time) -> String {
+    git::gix_time_to_time(time)
+        .format(&time::format_description::well_known::Rfc3339)
```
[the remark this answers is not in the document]
Shnatsel: Yup, that worked.

rustsec/src/repository/git/commit.rs
Byron: This is now available via [this PR](https://github.com/Byron/gitoxide/pull/973) will be called `commit.signature()`. Basic validation seems straightforward if `gpg` can be used as well.

rustsec/src/repository/git/repository.rs
```diff
@@ -72,64 +69,99 @@ impl Repository {
...
+            path.with_extension("rustsec"),
+            gix::lock::acquire::Fail::AfterDurationWithBackoff(std::time::Duration::from_secs(
+                60 * 10, /* 10 minutes */
+            )),
+            Some(std::path::PathBuf::from_iter(Some(
+                std::path::Component::RootDir,
+            ))),
+        )
+        .map_err(|err| format_err!(ErrorKind::Repo, "unable to acquire repo lock: {}", err))?;
+
```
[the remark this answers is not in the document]
Byron: Please note that unless [signal handlers are setup](https://docs.rs/gix/0.51.0/gix/interrupt/fn.init_handler.html), it's likely that users will one day hit `Ctrl + C` during the operation, which leaves the lock file stranded. With handlers, locks will be removed in such an event as well.

Now that `gix` *can* use `tracing` if compiled in, I'd also consider updating `gix_lock` to emit trace messages if the waiting time goes past a certain point so people have a chance to see which file it blocks on.

rustsec/src/repository/git/repository.rs
Shnatsel: Keeping it as-is is a good call. Relying on `FETCH_HEAD` would introduce issues around relying on the client time - e.g. if the client time is set into the far future at the time of checkout, the DB would never be re-downloaded again.

There are of course possible mitigations to this but they add a lot more complexity.

Byron (commented):
Thanks for your help @Jake-Shadle and for letting me take a look 🙏.
My comments here a merely a quick skim over all existing comments, but I will be happy to do a review once you think it's ready and want me to take another look.

Shnatsel:
@Jake-Shadle is moving on to other things, so this *is* the final version of this PR.

I am going to merge this, and then fix it up based on your feedback @Byron once it's in master. You would be able to directly propose changes in a PR too.

PR #927, opened 2023-08-08
First fixup pass on the libgit2 -> gix transition

Shnatsel:
Builds upon #925


Shnatsel:
My local `rustfmt` and the one on CI apparently disagree about how a comment should be formatted.

PR #928, opened 2023-08-08
Faster CI for `cargo audit` tests

Shnatsel:
Speeds up CI by:

1. Running `cargo test` without the `--release` flag, avoiding compiling the code twice (tests are run with `cargo run` without `--release`)
1. Using the sparse index instead of fetching crates.io index via git, which would cause a regression in test performance (`gix` and its deps such as `miniz_oxide` compiled in debug mode)

The Rust toolchain is bumped to v1.71 for sparse index support but MSRV is still enforced by the `workspace` job.

### DIFF
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -47,7 +47,7 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           profile: minimal
           override: true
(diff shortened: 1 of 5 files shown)


Shnatsel:
`rustsec / test` on MSRV down from 14m 36s to 8m 6s

PR #929, opened 2023-08-09
Add Rust-specific caching job to see if that speeds up CI

Shnatsel:


tarcieri:
I guess we can close #551 now

PR #934, opened 2023-08-09
Less surprising locking

Shnatsel:
Polish on top of #925

### DIFF
--- a/rustsec/src/repository/git/repository.rs
+++ b/rustsec/src/repository/git/repository.rs
@@ -69,19 +69,37 @@ impl Repository {
         if path.is_dir() && fs::read_dir(&path)?.next().is_none() {
             fs::remove_dir(&path)?;
         }
+
         // Set up signal handlers so that the lock is released if the user presses Ctrl+C
         // see https://github.com/rustsec/rustsec/pull/925#discussion_r1287265212
         gix::interrupt::init_handler(||{}).unwrap();
-        let _lock = gix::lock::Marker::acquire_to_hold_resource(
-            path.with_extension("rustsec"),
-            gix::lock::acquire::Fail::AfterDurationWithBackoff(std::time::Duration::from_secs(
-                60 * 10, /* 10 minutes */
-            )),
-            Some(std::path::PathBuf::from_iter(Some(
+
+        // Lock the directory with the git repo checkout so that several processes don't trample each other
+        const LOCK_WAIT_MINUTES: u64 = 10;
+        let _lock = {
+            let boundary_dir = Some(std::path::PathBuf::from_iter(Some(
                 std::path::Component::RootDir,
-            ))),
-        )
-        .map_err(|err| format_err!(ErrorKind::Repo, "unable to acquire repo lock: {}", err))?;
+            )));
+            // Attempt to acquire the lock without waiting. If this fails, print a message and wait.
+            // This is done to avoid `cargo audit` "hanging" without a clearly communicated reason.
+            match gix::lock::Marker::acquire_to_hold_resource(
+                path.with_extension("rustsec"),
+                gix::lock::acquire::Fail::Immediately,
+                boundary_dir.clone()) {
+                    Ok(marker) => marker,
+                    Err(e) => {
+                        println!("Could not acquire the lock on git repository at {path:?}: {e}. Waiting for up to {LOCK_WAIT_MINUTES} minutes to acquire the lock.");
+                        gix::lock::Marker::acquire_to_hold_resource(
+                            path.with_extension("rustsec"),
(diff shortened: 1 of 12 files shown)


rustsec/src/repository/git/repository.rs
```diff
@@ -69,16 +69,40 @@ impl Repository {
...
-        let _lock = gix::lock::Marker::acquire_to_hold_resource(
-            path.with_extension("rustsec"),
-            gix::lock::acquire::Fail::AfterDurationWithBackoff(std::time::Duration::from_secs(
-                60 * 10, /* 10 minutes */
-            )),
-            Some(std::path::PathBuf::from_iter(Some(
+
+        // Set up signal handlers so that the lock is released if the user presses Ctrl+C
+        // see https://github.com/rustsec/rustsec/pull/925#discussion_r1287265212
+        gix::interrupt::init_handler(|| {}).unwrap();
```
Shnatsel, line 75: @Byron is `gix::interrupt::init_handler` safe to call multiple times?

Byron: Yes, even though it will [return an error then](https://github.com/Byron/gitoxide/blob/c6785fc7082b90c8a27cef6a0f5cc5acd8cb8951/gix/src/interrupt.rs#L71-L76). I don't know if this is the best behaviour though, maybe it should just silently skip re-initialization?

Shnatsel: Also, I tried running `Deregister` in the `Drop` implementation for `Repository` but the gix type `Deregister` is not properly public and cannot be referred to.

Shnatsel: Oh, so the handler can be registered only once globally and returns an error for subsequent initialization?

This is really inconvenient because I need to register the handler every time locking is attempted, but attempts at locking may run concurrently from several threads (which is kind of the whole point of locking). So if the first thread registers a handler, locks, does the work, unregisters the handler at the end and quits, the second thread that now acquires the lock is running without the signal handler and will leave a stale lockfile if it receives SIGINT.

What is needed here is atomic reference counting: every time `init_handler` is called the count is incremented by one, every time `deregister` is called the count is decremented. Once the count reaches 0 the signal is actually deregistered. I think this needs to be implemented in `gix`, doing this client-side would be rather hacky.

Byron: I have implemented this here: https://github.com/Byron/gitoxide/pull/975, and it should fulfil the requirements.

Shnatsel: @Byron how confident are you that the signal handlers are actually needed?

I experimented with the `main` branch, interrupting the process during crates.io checkout with the signal handlers never registered from `rustsec` code, and I could not get any stale locks to impede subsequent runs.

Shnatsel: I've also checked that the signal handlers are not already set by something else - afterwards I added `dbg!(gix::interrupt::init_handler(|| {}).is_ok());` which shows `true`, and since I'm using the stable `gitoxide` in `main` branch, this means that the handler is not already set.

Byron: > @Byron how confident are you that the signal handlers are actually needed?

Destructors are typically run if `Ctrl + C` is pressed, which also causes lock files to be cleaned up. If the binary is compiled with `panic = abort`, handler are required, otherwise they are indeed *quite* optional. The problem is when the program is busy, then *I think* it will also just be aborted.

I just checked and realized that `gix` is also compiled with `panic = "unwind"` even though a comment above says that signal handlers would be preferred as the binaries will get smaller that way, so I think I will change that :).



> I've also checked that the signal handlers are not already set by something else - afterwards I added `dbg!(gix::interrupt::init_handler(|| {}).is_ok());` which shows `true`,

This probably means that this function isn't actually called multiple times by multiple threads, but only one at a time. That's not entirely surprising, as having multiple threads in the same program race into this code seemed quite unlikely in the first place. *It's good to be thread-safe now though, and the `init_handler()` function was improved substantially thanks to you🙏*.

Shnatsel: Thank you for all your help with this transition!

Shnatsel: With `rustsec` being a library crate, I have chosen to document the locking considerations instead of registering a signal handler behind the scenes: https://github.com/rustsec/rustsec/pull/935

This should give the API consumers more control. We can also make the timeout configurable later, but I'll wait for someone to with a use case for it to show up before over-engineering the library.

Byron: I think that's the right choice! Until know I thought we were in an application, which is why I was so easy about it🤦‍♂️.

PR #937, opened 2023-08-10
Foolproof API for the cached index

Shnatsel:
Cleaning up the APIs after the addition of sparse index support in #923:

 - Make `find_yanked()` automatically populate the cache
 - Make `is_yanked()` private because of the performance pitfall of calling it in a loop without populating cache first.
   - With it gone, there is no reason to expose `populate_cache()` publicly, so make it private too.
 - Adjust `cargo-audit` code accordingly
 - Slightly improve documentation

@Jake-Shadle FYI


Shnatsel:
I am going to go ahead and merge this because having this in a branch is blocking further work.

PR #955, opened 2023-08-23
CC-BY attribution in `cargo audit`

Shnatsel:


Shnatsel:
This is pretty verbose but the alternatives that don't duplicate the print statement are [even worse](https://github.com/rustsec/rustsec/pull/955/commits/b5f660727289e21f8bbe22e961e1ebe9160ccfdb).

cargo-audit/src/presenter.rs
```diff
@@ -307,10 +308,21 @@ impl Presenter {
...
         self.print_attr(color, "ID:       ", &metadata.id);
 
-        if let Some(url) = metadata.id.url() {
-            self.print_attr(color, "URL:      ", url);
-        } else if let Some(url) = &metadata.url {
-            self.print_attr(color, "URL:      ", url);
+        if metadata.license == License::CcBy40 {
+            // We must preserve the original URL from the `url` field
+            if let Some(url) = &metadata.url {
+                self.print_attr(color, "URL:      ", url);
```
amousset, line 314: Hum. I'm wondering if we shouldn't display the RustSec URL anyway (in addition to the source URL) as the advisory metadata might be different on GitHub side (in case we have re-qualified it), and might not match what is displayed in the output.

Shnatsel: This is another question of policy. Do we always forward the advisories unaltered, and apply changes upstream? (There is now a Github repo where you can make changes in OSV format via PRs).

Or do we make changes on our side and assign our own IDs to the altered advisory?

`rustsec` crate already supports `GHSA-*` advisory IDs as it turns out. If we always use GHSA- advisories in our DB then special handling for the `url` field isn't even necessary.

Shnatsel: Also, a big part of the motivation is to make all complaints about the advisory contents to go to Github. If it isn't our data it isn't our problem :stuck_out_tongue_closed_eyes: 

Of course we would still review incoming advisories for sanity but in case of any disputes they would happen upstream.

tarcieri: Using `GHSA-*` identifiers for advisories we import there seems good to me (versus giving them RUSTSEC ones)

Shnatsel: My proposal for handling GHSA data is to import advisories as-is, and make any changes we deem necessary upstream, mostly so that we don't create 2 different contradictory advisories, which is very confusing for security responders.

In this situation it makes sense to reuse GHSA-* IDs. If we were altering the data it would make sense to assign a RUSTSEC ID, but if we don't alter it then it's best to use the single canonical identifier.

Shnatsel: Yay or nay?

tarcieri: I'd suggest not altering imported advisories on our side, but doing any alterations upstream and then re-syncing.

Shnatsel: Sounds good to me. 

The code in this PR already supports that. Should I go ahead and merge this?

tarcieri (approved):

PR #983, opened 2023-09-01
make 'cargo test --no-default-features' run without errors

alexanderkjall:
Without this I got errors like this one:

```
error[E0433]: failed to resolve: could not find `v3` in `cvss`
  --> cvss/tests/base.rs:11:22
   |
11 |     let base = cvss::v3::Base::from_str(cvss_for_cve_2013_1937).unwrap();
   |                      ^^ could not find `v3` in `cvss`
```

Ref: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/cvss/debian/patches/fix-unit-tests.patch

### DIFF
--- a/cvss/tests/base.rs
+++ b/cvss/tests/base.rs
@@ -5,7 +5,7 @@
 use core::str::FromStr;
 
 /// CVE-2013-1937
-#[cfg(feature = "default")]
+#[cfg(all(feature = "v3", feature = "std"))]
 #[test]
 fn cve_2013_1937() {
     let cvss_for_cve_2013_1937 = "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N";
@@ -15,7 +15,7 @@ fn cve_2013_1937() {
 }
 
 /// Missing CVSS:3.1 prefix
-#[cfg(feature = "default")]
+#[cfg(all(feature = "v3", feature = "std"))]
 #[test]
 fn bad_prefix() {
     let cvss_for_cve_2013_1937e = "AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N";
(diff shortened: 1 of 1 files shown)


cvss/tests/base.rs
```diff
@@ -5,6 +5,7 @@
 use core::str::FromStr;
 
 /// CVE-2013-1937
+#[cfg(feature = "default")]
```
tarcieri, line 8: These should all be...

```suggestion
#[cfg(feature = "v3")]
```

...but other than that this looks good, thanks

tarcieri: Note my comment only fixed one of them: there are several more that should all be `v3`

alexanderkjall: This will case an error of:
```
error[E0599]: no method named `score` found for struct `Base` in the current scope
   --> cvss/tests/base.rs:274:21
    |
274 |     assert_eq!(base.score().value(), 1.8);
    |                     ^^^^^ method not found in `Base`
```

But I'll set them to `v3` and `std`

alexanderkjall:
Thanks for the review and sorry for the force push, but that seemed like the cleanest way to apply to feedback :)

Shnatsel:
Let's go with the `#![cfg` approach. If we add tests that don't require it, we'll just add them to another file.

alexanderkjall:
fixed :)

Shnatsel:
Thanks!

Shnatsel:
@alexanderkjall btw you might want to start by packaging `cargo-audit` v0.17.6, because v0.18.x has a lot more dependencies some of which are evolving rapidly (e.g. gitoxide) and also only supports x86 and ARM, which may be a deal-breaker for Debian.

alexanderkjall:
That does indeed sound like a pain point, it's not a total dealbreaker. We already have some software that is crippled like that, for example ring and everything that depends on it, but it complicates the packaging work a fair bit. I'll start with 0.17.6.

PR #987, opened 2023-09-05
Do not re-lookup packages that are already cached

Shnatsel:
Fixes a performance regression when scanning multiple files. Only possible in `cargo audit bin`.


Shnatsel:
No change in output on the contents of my `~/.cargo/bin/`, so I am confident this is correct.

PR #995, opened 2023-09-06
Fix `deny = ["warnings"]` being ignored

### LINKED ISSUE #991: Incosistency between config file and CLI
If I use `cargo audit --deny warnings` it throws an error on yanked crates.

But if I use:
```toml
[output]
deny = ["warnings"]
quiet = false
```
it won't. It will continue to only throw a warning.

Curiously if I use `deny = ["yanked"]` it will correctly throw an error on yanked crates.

Used v0.18.1.

daxpedda:
I'm not sure why `filter_map()` was used in the first place, but that fixes it.

Initially I was looking into adding tests, but unfortunately `abscissa_core::testing::CmdRunner` doesn't support settings the working directory, so I'm unsure how. Happy to take any directions though.

Fixes #991.

### DIFF
--- a/cargo-audit/src/config.rs
+++ b/cargo-audit/src/config.rs
@@ -189,12 +189,16 @@ impl DenyOption {
         ]
     }
     /// Get the warning::Kind that corresponds to self, if applicable
-    pub fn get_warning_kind(self) -> Option<WarningKind> {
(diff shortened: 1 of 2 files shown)


Shnatsel:
I'm not familiar with this part of the codebase. @tarcieri I'd appreciate if you could take a look.

tarcieri (approved):

PR #1017, opened 2023-09-25
rustsec-admin with tame-index

Byron:
This PR switches rustsec-admin to tame-index, which uses `gix`.

Originally the idea was to speed the linting which takes 3 minutes on CI, 
but as it turns out the linting itself takes only about 1s.

What can take time though is the initial clone of the crates-index. Using git2, this
was single-threaded and took about 1m50s. Now, with a single thread, it's done in 47s. With all 10 cores it's at 24s (most of which is downloading the pack).

I also enabled multi-threading and performance options to speed it up even more.
Unfortunately I don't know how effective these changes are on CI when linting advisories though.

### Tasks

* [x] switch to `tame-index`
* [x] allow multi-threading when dealing with git and enable performance options

### Review Notes

I now see…

```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

…when compiling and don't know if this is due to setting edition 2021 in `rustsec-admin`. It would be easy to drop the edition change.

* Should the clippy errors by fixed by boxing, or just be ignored?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,6 +93,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+
+[[package]]
 name = "askama"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -478,6 +484,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b8191fa7302e03607ff0e237d4246cc043ff5b3cb9409d995172ba3bea16b807"
 
 [[package]]
+name = "cmake"
+version = "0.1.50"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a31c789563b815f77f4250caee12365734369f942439b7defd71e18a48197130"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "color-eyre"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


Shnatsel:
Thanks a lot! It is great to drops OpenSSL from the dependency tree on CI.

Multi-threading and performance options are not required. Instead it would be more efficient to switch to using sparse index in place of git index. I assume using sparse index in a performant manner will require a bit of refactoring, so that's out of scope for this PR.

Shnatsel (approved):

Byron:
The multi-threading is just happening when cloning the crates-index, nothing else was modified. I think it's a cheap way to speed up CI for advisories, as like you said, sparse index support requires more work.
Just to verify: do you really want to only use 1 core for pack resolution instead of the maximum amount available on the CI agent? On a 10-core machine the crates-io-clone + lint workload now takes 24s, down from 47s from single-threaded operation, and down from 1m50s when using `git2`.

Regrading `clippy`, I don't get it as I have fixed these warnings and can't reproduce them locally.

Shnatsel:
The Clippy warnings are my fault, and also happen in the main branch. You can safely ignore them in this PR.

I'm fine with using more cores, I just don't want to deal with memory safety vulnerabilities in C code on CI that has the secret token for our repo. `max-performance-safe` would be preferable.

Shnatsel:
By the way, instead of `-for-configuration-only` crates you can use the new `dep?/feature` syntax, see https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

So it would go something like this:
```
[features]
default = ["gix?/max-performance-safe"]
```

That's not blocking this PR though, I can refactor it later myself no problem.

Byron:
> By the way, instead of `-for-configuration-only` crates you can use the new `dep?/feature` syntax, see https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

That would be *wonderful*!! The way I was using was always like a hack. Let me try it.

Regarding clippy, I usually ignore these kind of warnings (Err variant too big),  but maybe better leave it to you to decide.

Byron:
Unfortunately, the proposed `[features]` style for configuration only works if `gix` is an optional dependency - I thought it would allow to avoid pulling in `gix` as direct dependency.
Probably that's still better though.

With this I'd be at the end of what I think I can do.

Shnatsel:
Thanks a lot!

Shnatsel:
This is now live in production. Lint times on CI are down from 2m41s to 1m14s. Thanks!

tarcieri:
Did this also indirectly fix the release action?

Shnatsel:
Probably, yeah. I have not tested it though.

PR #1018, opened 2023-09-25
Cargo audit release with a fix for RUSTSEC-2023-0064

Shnatsel:
Release a new `cargo audit` version so that Linux distributions would pick it up, and people who installed it through `cargo` would get a fix with a `cargo install` instead of requiring `cargo install --force`.

I think this is a good idea, but I'd like one of the other maintainers to sign off on it before I ship it.


amousset (approved):
Makes sense.

PR #1032, opened 2023-09-30
Switch from Git-compatible locks to OS locks

### LINKED ISSUE #1011: Stale lockfile encountered in the wild
Reported at https://github.com/Byron/gitoxide/issues/1026

Shnatsel:
Switch from Git-compatible locks provided by `gix` to an interface around OS locks provided by `tame-index`. Git locks are very poorly designed, being plain lock files without even a PID written into them; so stale locks are left behind on SIGKILL or power loss with no way to recover.

Also acquire the global Cargo package lock when doing crates.io index lookups, to avoid racing with Cargo itself when fetching crates index.

Fixes #1011

### DIFF
--- a/admin/src/lock.rs
+++ b/admin/src/lock.rs
@@ -1,0 +1,24 @@
+//! Utiltity functions for file locking
+
+use std::time::Duration;
+
+use tame_index::{utils::flock::LockOptions, index::FileLock};
+
+/// Acquires the Cargo package lock, or fails immediately
+pub fn acquire_cargo_package_lock() -> Result<FileLock, tame_index::Error> {
+    let lock_opts = LockOptions::cargo_package_lock(None)?.exclusive(false);
+    acquire_lock(lock_opts, Duration::from_secs(0))
+}
+
+/// Acquires the provided lock with a speicifed timeout
+pub fn acquire_lock(
+    lock_opts: LockOptions<'_>,
+    lock_timeout: Duration,
+) -> Result<FileLock, tame_index::Error> {
+    if lock_timeout == Duration::from_secs(0) {
+        lock_opts.try_lock()
+    } else {
+        lock_opts.lock(|_| Some(lock_timeout))
+    }
+}
+
--- a/admin/src/lock.rs
+++ b/admin/src/lock.rs
@@ -10,7 +10,7 @@ pub fn acquire_cargo_package_lock() -> Result<FileLock, tame_index::Error> {
     acquire_lock(lock_opts, Duration::from_secs(0))
 }
 
-/// Acquires the provided lock with a speicifed timeout
+/// Acquires the provided lock with a specified timeout
 pub fn acquire_lock(
     lock_opts: LockOptions<'_>,
     lock_timeout: Duration,
--- a/admin/src/lock.rs
+++ b/admin/src/lock.rs
@@ -1,4 +1,4 @@
-//! Utiltity functions for file locking
+//! Utility functions for file locking
 
 use std::time::Duration;
 
--- a/admin/src/lock.rs
+++ b/admin/src/lock.rs
@@ -2,7 +2,7 @@
 
 use std::time::Duration;
 
-use tame_index::{utils::flock::LockOptions, index::FileLock};
+use tame_index::{index::FileLock, utils::flock::LockOptions};
 
 /// Acquires the Cargo package lock, or fails immediately
 pub fn acquire_cargo_package_lock() -> Result<FileLock, tame_index::Error> {
@@ -21,4 +21,3 @@ pub fn acquire_lock(
         lock_opts.lock(|_| Some(lock_timeout))
     }
 }
-
--- a/rustsec/src/repository/git/repository.rs
+++ b/rustsec/src/repository/git/repository.rs
@@ -99,13 +99,15 @@ impl Repository {
         // Lock the directory to avoid several checkouts running at the same time trampling on each other.
         // We do not use Git locks because they are very poorly designed - they leave stale locks on SIGKILL or power loss
         // with no way to recover. They don't even write the PID to the lockfile.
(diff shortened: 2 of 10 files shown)


Shnatsel:
@Jake-Shadle is `acquire_cargo_package_lock()` appropriate when I'm just fetching the index? Or is it for caching actual .crate files only?

This PR is written assuming that the package index operations will need a lock, but this is not described in the docs for `acquire_cargo_package_lock()` so I thought I'd double-check.

Jake-Shadle:
Oh sorry that is not clear, the package lock, as used by cargo, basically encompasses all disk i/o done on $CARGO_HOME which includes the index, crate cache, crate sources etc

Shnatsel:
Cool, thanks for clarifying!

Then this is ready for review.

Byron:
> Git locks are very poorly designed, being plain lock files without even a PID written into them; so stale locks are left behind on SIGKILL or power loss with no way to recover.

Reading this, I find it worth noting that I keep forgetting the reason *why* `git` locks are the way they are, and I think it was the second time I re-discovered these as issues with these locks inevitably appear.

* these locks are meant to work on all filesystems (or at least the largest possible set of them)
* they are primarily used for writing files safely as in: swap the final version of the written resource in atomically, making half-written corrupt files impossible.
* **keep many resources locked at the same time**

The last point is the primary reason, to me, why `git` truly had to decide what to use locks as it does: the ref-database for instance may create a lot of locks at the same time if packed-refs aren't used, and as it stands, these are just files that *don't* have to be open. On the server, I am sure this increases the possible parallelism quite a bit as contention can happen with per-ref granularity (of course, at this point it's just a guess due to lack of experience with busy servers).

I hope you find this interesting, and even if not I can hope that I won't forget this another time 😁. And of course, please feel free to add corrections as all of the above is only my best guess, thus far I didn't stumble over the true historical reason for why `git` locks are what they are.

Shnatsel:
All of the points you list can be achieved with OS locks. In fact, OS locks would be basically useless if they didn't provide mechanisms for doing these things!

Simply replacing all calls to `open()` for Git lockfiles with `flock()` would be a tremendous improvement with pretty much no downsides. (It wouldn't work correctly on NFS, but git's lockfiles don't correctly on NFS either; NFS is just broken by design.)

The only thing Git gains from not using OS locking is not being OS dependent. Since git is written is C, they cannot simply pull in a library that provides cross-platform locking, they would have to reinvent it from scratch for all supported operating systems. My guess is that they didn't want to do that.

Byron:
I think there is a key-point where I understand the function of `flock` differently than you do: From what I can tell, these locks are only locking things if you have a file descriptor, and *I assume the lock goes away once you close it*. If that's not the case, then I agree `flock` can be used and then `gix-lock` could potentially learn to do that instead, maybe by adding a different kind of lock, so we'd have `git`-compatible ones and those who are not but that have other benefits. Even if the lock is still held after closing the handle (and as long as the owning process is alive), I wonder if there are limits to how many can be held at a time.

It would also be the question of such a new `gix-lock` type would have any advantage or semantic difference to what @Jake-Shadle would come up with in a separate crate though, I am just thinking in the direction of "could `gix-lock` be better" with or without losing `git` lock compatibility, and what it would all mean.

Shnatsel:
Ah, Linux does tend to have have a default limit of 1024 file descriptors open per process. If a process needs to hold a very large number of locks at the same time, this design is somewhat understandable. At least it gets from "obviously wrong" territory to "there are trade-offs".

Shnatsel:
@amousset would you like to sign off on the admin part, or should I just YOLO it?

rustsec/src/repository/git/repository.rs
```diff
@@ -104,20 +96,24 @@ impl Repository {
             fs::remove_dir(&path)?;
         }
 
-        // Acquire the lock on the checkout directory
-        let lock_policy = if lock_timeout == Duration::from_secs(0) {
-            gix::lock::acquire::Fail::Immediately
+        // Lock the directory to avoid several checkouts running at the same time trampling on each other.
+        // We do not use Git locks because they are very poorly designed - they leave stale locks on SIGKILL or power loss
```
Byron, line 100:
```suggestion
        // We do not use Git locks because they have undesirable properties - they leave stale locks on SIGKILL or power loss
```

rustsec/src/repository/git/repository.rs
```diff
@@ -104,20 +96,24 @@ impl Repository {
...
+        // with no way to recover. They don't even write the PID to the lockfile.
```
Byron, line 101:
```suggestion
        // with no way to recover, particularly because the PID isn't written to the lockfile.
```

Byron (commented):
I made a couple of suggestions with the intend to use more neutral language in a comment explaining why `git` lock files can't be used.

Otherwise it's lovely to see movement in this space as `cargo`-style file locks are certainly something that many applications can benefit from.

admin/src/lock.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use tame_index::{index::FileLock, utils::flock::LockOptions};
+
+/// Acquires the Cargo package lock, or fails immediately
+pub fn acquire_cargo_package_lock() -> Result<FileLock, tame_index::Error> {
+    let lock_opts = LockOptions::cargo_package_lock(None)?.exclusive(false);
+    acquire_lock(lock_opts, Duration::from_secs(0))
+}
+
+/// Acquires the provided lock with a speicifed timeout
```
amousset, line 13:
```suggestion
/// Acquires the provided lock with a specified timeout
```

admin/src/lock.rs
```diff
@@ -0,0 +1,23 @@
+//! Utiltity functions for file locking
```
amousset, line 1:
```suggestion
//! Utility functions for file locking
```

amousset (approved):

PR #1051, opened 2023-10-24
Fix deadlock on missing lockfile

### LINKED ISSUE #1050: cargo-audit hangs with "Blocking waiting for file lock on package cache"
If I clone [this repo](https://github.com/softdevteam/cactus) afresh, and run `cargo audit`:

```
$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 575 security advisories (from /home/vext01/.cargo/advisory-db)
    Updating crates.io index
    Blocking waiting for file lock on package cache
```

And it sits like that indefinitely...

There is no other cargo process running for the repo. No LSP or anything like that. I can reproduce this on both Debian and OpenBSD.

Using `strace` I see:
```
statx(AT_FDCWD, "Cargo.lock", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7ffe061f5360) = -1 ENOENT (No such file or directory)
```

Which got me thinking about `Cargo.lock`... If I run `cargo generate-lockfile` first, then `cargo audit` doesn't block.

So if cargo-audit requires a `Cargo.lock`, then should it make one automatically, or if it can't at least exit with a message like "please run `cargo generate-lockfile`"?

Shnatsel:
Fixes #1050

Sadly no tests because `abscissa::Runner` does not let you set the working directory, and adding it would require not only a new Abscissa release but also a migration from Abscissa v3 to v4 in `cargo-audit`, which is non-trivial.

### DIFF
--- a/cargo-audit/src/auditor.rs
+++ b/cargo-audit/src/auditor.rs
@@ -1,8 +1,6 @@
 //! Core auditing functionality
 
-use crate::{
-    binary_format::BinaryFormat, config::AuditConfig, lockfile, prelude::*, presenter::Presenter,
-};
+use crate::{binary_format::BinaryFormat, config::AuditConfig, prelude::*, presenter::Presenter};
 use rustsec::{registry, report, Error, ErrorKind, Lockfile, Warning, WarningKind};
 use std::{
     io::{self, Read},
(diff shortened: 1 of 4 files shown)


tarcieri:
Note: the only hard part of an `abscissa` upgrade is actually the associated `clap` v3 to v4 upgrade

Issue #503, opened 2022-01-07
cargo-audit unconditionally uses color escape codes when `--color=auto`


tmfink:
`cargo-audit` always uses ANSI color escape codes, even when `--color=auto`.

## Actual Behavior 

```
$ cargo r audit --color=auto | file -
...
/dev/stdin: UTF-8 Unicode text, with escape sequences

$ cargo r audit | file -
...
/dev/stdin: UTF-8 Unicode text, with escape sequences
```

It seems that the `NO_COLOR` environment variable *is* respected:
```
$ NO_COLOR=1 cargo r audit | file -
...
/dev/stdin: UTF-8 Unicode text
```

## Desired Behavior

When `--color=auto`, ANSI escape codes should only be used when writing directly to a console (not when output redirected to a file or piped to another program).

tarcieri:
Bumping to `clap` v4 will get rid of the `atty` dependency and replace it with `is-terminal`

qtfkwk:
https://github.com/rustsec/rustsec/pull/1057

PR #1057, opened 2023-11-05
fix https://github.com/rustsec/rustsec/issues/503

qtfkwk:


Shnatsel (changes requested):
There is actually a subtlety to it - the stderr may be a terminal while stdout isn't, or vice versa. Right now this only checks stdout; stderr should be taken into account also.

Ideally we'd auto-detect the color for stdout and stderr separately, and write to each one appropriately; but I'm fine with falling back to plain text output if either of them is not a terminal, as a starting point.

qtfkwk:
I'm not familiar with abscissa... but it seems that configuring term colors for stdout and stderr independently isn't supported (?)... and complicated by abscissa_core::terminal::streams::Streams apparently handled together (?), whereas something else like bunt has [`set_stderr_color_choice`](https://docs.rs/bunt/latest/bunt/fn.set_stderr_color_choice.html) and [`set_stdout_color_choice`](https://docs.rs/bunt/latest/bunt/fn.set_stdout_color_choice.html).

I've gone ahead and modified it to check both and make the default be `Never` if either stdout or stderr is not a terminal as you suggested for a better starting point.

Shnatsel:
Thanks!

PR #1044, opened 2023-10-14
`cargo update`

Shnatsel:
Updates dependencies to latest semver-compatible versions.

PR #1062, opened 2023-11-16
cargo update

Shnatsel:
Updates dependencies to latest semver-compatible versions.

Bumps `cargo-lock` msrv to 1.67.

Supersedes #1044


tarcieri:
Perhaps it's time to reconsider our MSRV policies

Issue #1029, opened 2023-09-28
cargo-audit fails to fetch advisory database: "Could initialize the http client"


kpreid:
I just decided to upgrade my `cargo-audit` from `v0.17.6` to `v0.18.2`. The command then failed:

```
$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
error: couldn't fetch advisory database: git operation failed: failed to prepare fetch: Could initialize the http client
```

I thought that perhaps there was a problem with the existing advisory-db, so I deleted ~/.cargo/advisory-db. Now it fails with:

```
$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
error: couldn't fetch advisory database: git operation failed: failed to fetch repo: Could initialize the http client
```

`cargo audit --no-fetch` succeeds, given a manually updated `advisory-db`. `cargo install --locked cargo-audit` did not make a difference to the behavior.

I'm assuming this is a bug, but if it's some misconfiguration then it would be nice if the error message explained what to do about it (and wasn't missing a "not" after "Could").

* OS: macOS 12.7 (21G816)
* rustc 1.72.1 (d5c2e9c34 2023-09-13)
* x86_64-apple-darwin

Shnatsel:
The error message missing a "not" originates from the `gix` crate, and the message seems to be cut off also due to `gix` behavior. I have filed an issue upstream: https://github.com/Byron/gitoxide/issues/1075

Once that's fixed it will at least start reporting the underlying problem correctly, so we will be able to debug it.

Byron:
The lack of detail is due to the [`source()` trail](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) not being followed by the [routine that prints the error](https://github.com/RustSec/rustsec/blob/3544515990b09441ecc12df8d0291bc6f23d3d30/cargo-audit/src/auditor.rs#L73). `Display` will only display the error message of the respective error, but the error provided by `cargo-audit` seems to be made to concatenate sources of errors by hand like `msg: {source}` and expects the error in `source` to do the same. `gix` errors don't do that as it cannot make a choice for the user (in terms of error display and presentation) beyond the human-readable message that presents the context of the error at hand.

One could add a custom printer to additionally follow the error chain via `source()` and keep appending to the line to add the desired context.

Shnatsel:
[...]

PR #1067, opened 2023-11-18
Add a `source` field to `rustsec::Error`, and use it in simple cases.

kpreid:
As discussed in https://github.com/rustsec/rustsec/issues/1029#issuecomment-1815785981

This is a semver-compatible change because it only adds a private field and a new constructor: `rustsec::Error::with_source()`.

Note that there are many remaining cases where `Error` is constructed with `format_err!`, that stringify a source error; this PR only addresses the easy cases of the `Error::new()` calls.

### DIFF
--- a/rustsec/src/error.rs
+++ b/rustsec/src/error.rs
@@ -39,16 +39,48 @@ pub struct Error {
     /// Kind of error
     kind: ErrorKind,
 
-    /// Message providing additional information
+    /// Message providing a more specific explanation than `self.kind`.
+    ///
+    /// This may be a complete error by itself, or it may provide context for `self.source`.
     msg: String,
+
+    /// Cause of this error.
+    ///
+    /// The specific type of this error should not be considered part of the stable interface of
+    /// this crate.
+    source: Option<Box<dyn std::error::Error + Send + Sync>>,
 }
 
 impl Error {
-    /// Create a new error with the given description
+    /// Creates a new [`Error`](struct@Error) with the given description.
+    ///
+    /// Do not use this for wrapping [`std::error::Error`]; use [`Error::with_source()`] instead.
     pub fn new<S: ToString>(kind: ErrorKind, description: &S) -> Self {
+        // TODO: In a semver-breaking release, deprecate accepting anything but a `String`,
+        // or maybe `AsRef<str>`. This will discourage putting error types in the `description`
+        // position, which makes it impossible to retrieve their `.source()` info. It will also
+        // avoid an unnecessary clone in the common case where `S` is already a `String`.
         Self {
             kind,
             msg: description.to_string(),
+            source: None,
+        }
+    }
+
+    /// Creates a new [`Error`](struct@Error) whose [`std::error::Error::source()`] is `source`.
+    ///
+    /// `msg` should describe the operation which failed so as to give context for how `source`
+    /// is a meaningful error. For example, if `source` is a [`std::io::Error`] from trying to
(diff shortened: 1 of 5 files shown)


rustsec/src/error.rs
```diff
@@ -39,16 +40,48 @@ pub struct Error {
...
+            source: None,
+        }
+    }
+
+    /// Creates a new [`Error`] whose [`std::error::Error::source()`] is `source`.
+    ///
+    /// `msg` should describe the operation which failed so as to give context for how `source`
+    /// is a meaningful error. For example, if `source` is a [`std::io::Error`] from trying to
+    /// read a file, then `msg` should include the path of the file and why the file is relevant.
+    pub fn with_source<E: ErrorTrait + Send + Sync + 'static>(
```
tarcieri, line 76: I think this would be clearer without the renamed import:

```suggestion
    pub fn with_source<E: std::error::Error + Send + Sync + 'static>(
```

rustsec/src/error.rs
```diff
@@ -60,11 +93,27 @@ impl Error {
 
 impl Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}: {}", &self.kind, &self.msg)
+        let Self {
+            kind,
+            msg,
+            source: _,
+        } = self;
+        write!(f, "{kind}: {msg}")
```
tarcieri, line 101: Not sure this is an improvement. Maybe if you used a couple of local bindings rather than matching on the whole struct:

```suggestion
        let kind = self.kind;
        let msg = &self.msg;
        write!(f, "{kind}: {msg}")
```

rustsec/src/error.rs
```diff
@@ -1,6 +1,7 @@
 //! Error types used by this crate
 
 use std::{
+    error::Error as ErrorTrait,
```
tarcieri, line 4: To import this while avoiding a name collision you can do:

```suggestion
    error::Error as _,
```

rustsec/src/error.rs
kpreid: The purpose of the `Self {}` pattern is to force a compilation error if a field is added to the struct but this code isn't updated, to avoid mistakes where a field _should_ be included in the formatting but is not. If you disprefer that, I'd rather stick to no `let`s at all. Which shall I proceed with?

tarcieri: I think the original code is fine, personally

rustsec/src/error.rs
kpreid: Turns out importing the trait isn't needed at all, given that we're going to use it fully qualified.

rustsec/src/error.rs
```diff
@@ -60,11 +93,27 @@ impl Error {
...
+            kind,
+            msg,
+            source: _,
+        } = self;
+        write!(f, "{kind}: {msg}")
     }
 }
 
-impl std::error::Error for Error {}
+impl ErrorTrait for Error {
```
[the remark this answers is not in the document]
kpreid: Done.

rustsec/src/error.rs
kpreid: Done.

kpreid:
Sorry for forgetting to also address the clippy and rustdoc lints, but I believe it's all set now.

tarcieri (approved):

tarcieri:
Thanks!

PR #1081, opened 2023-12-30
gix upgrade to v0.56

Byron:
This is the answer to https://github.com/Byron/gitoxide/issues/1216#issuecomment-1872388629 .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -437,7 +437,7 @@ dependencies = [
  "android-tzdata",
  "iana-time-zone",
  "num-traits",
- "windows-targets",
+ "windows-targets 0.48.5",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


Byron:
That's unexpected: Some crates changed their MSRV because of a find-and-replace, and have been republished as part of the v0.57 release, but without a breaking change. This is why they are now picked up here and fail CI due to a lower MSRV.

Ideally I'd be able to determine the MSRV for all crates and keep the minimal required Rust version (or the `rust-version` field in the manifest) at exactly what it needs to be.

In theory, I can re-publish with a lower MSRV and make a patch release that fixes it, as effectively nothing changed when upping the MSRV to 1.70. Let me do that… .

Shnatsel:
We're not very strict about MSRV. If it causes a lot of hassle, I'd rather just bump it than deal with all the extra hassle.

Byron:
Thanks, that's good to know! This time I think we can keep it as is, but eventually the MSRV will updated to v1.70 once dependencies of `gitoxide` enforce it. Independently of that, the declared `rust-version` can be as low as it has to be as there is usually now benefit if it's increased without actually using the respective feature or API.

Shnatsel:
FWIW there's https://github.com/foresterre/cargo-msrv to automatically detect the MSRV of your crates. But knowing how many crates you have, it may take a while.

Shnatsel:
@Byron why is this marked as draft? Is there anything in particular that we're waiting on?

Byron:
I think this must be an oversight - sorry for the delay.

tarcieri:
Can we merge this?

PR #1068, opened 2023-11-19
[WIP] cargo-audit: bump abscissa to v0.7, clap to v4

tarcieri:
This compiles but the acceptance tests fail and I'm not sure why

PR #1092, opened 2024-02-01
Update abscissa_core and clap

therealprof:
This updates abscissa_core and clap to the current versions to get rid of a few outdated dependencies.

### DIFF
--- a/.github/workflows/cargo-lock.yml
+++ b/.github/workflows/cargo-lock.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.67.0 # MSRV
+          - 1.70.0 # MSRV
           - stable
     steps:
       - uses: actions/checkout@v4
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.67.0 # MSRV
+          - 1.70.0 # MSRV
           - stable
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/quitters.yml
+++ b/.github/workflows/quitters.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.67.0 # MSRV of `cargo audit`, should be bumped in tandem
+          - 1.70.0 # MSRV of `cargo audit`, should be bumped in tandem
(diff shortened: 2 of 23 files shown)


Shnatsel:
Thank you! This conversion is long overdue, and I would be very happy to see it happen.

I see some dependencies require rust 1.70, which causes CI to fail. So we also need to bump Rust toolchain version in https://github.com/rustsec/rustsec/tree/main/.github/workflows files and bump MSRV in `Cargo.toml`.

therealprof:
@Shnatsel I've bumped the MSRV in Cargo.tomls, GHA and READMEs. I wasn't sure what to do about the CHANGELOG since there doesn't seem to be a placeholder section for the next release as of yet. 🤞🏻

tarcieri:
Hmm, are the build failures related to #1081?

Shnatsel:
It seems that they are. Strange, I thought CI passed before I merged it? Anyway, I'll revert it so that this PR can be merged without gix weirdness getting in the way.

Shnatsel:
It seems there are genuine test failures in acceptance tests.

therealprof:
I'll have a look.

tarcieri:
Aah yeah, those were the some ones I was running into in #1068. Never figured out what the cause was. `clap` seems to be doing something weird.

therealprof:
I can totally believe that. I had tons of problems upgrading from clap 3 to clap 4 in the past.

therealprof:
Okay, here's the situation: I got most of the failures fixes (conflict between the automatically generated help flag and manually defined one -- don't ask me how that could produce those errors).

But:
```
thread 'version' panicked at cargo-audit/tests/acceptance.rs:207:5:
assertion `left == right` failed
  left: "cargo-audit-audit 0.18.3\n"
 right: "cargo-audit 0.18.3\n"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I can't for my life figure out how to override name in the version string of the `audit` subcommand. Probably would require some refactoring and it's starting to get late here, so I'm going to change the unittest for now.

tarcieri:
@therealprof works for me. We can followup on fixing that (if possible)

tarcieri (approved):

PR #1094, opened 2024-02-01
Revert "gix upgrade to v0.56"

Shnatsel:
Reverts rustsec/rustsec#1081

Seems like it caused build failures, so reverting to unblock #1092

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -437,7 +437,7 @@ dependencies = [
  "android-tzdata",
  "iana-time-zone",
  "num-traits",
- "windows-targets 0.48.5",
+ "windows-targets",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


tarcieri:
cc @Byron

Shnatsel:
Tests pass on main with gix 0.56 PR merged, so it's probably some interaction between https://github.com/rustsec/rustsec/pull/1092 and https://github.com/rustsec/rustsec/pull/1081

Byron:
Thanks for reeling me in. In case it's something that can be fixed in `gix`, maybe you can share the build failures that are referred to?

Shnatsel:
I've sorted it out in #1099, all good. Thanks again for your help @Byron !

PR #1101, opened 2024-02-03
Release `cargo audit` 0.18.4

Shnatsel:


tarcieri:
Perhaps with all of the changes including MSRV it should be 0.19

Shnatsel:
Did we actually change anything in the CLI? If so, I agree. Otherwise I'm inclined to keep it in the 0.18.x series.

tarcieri:
It's a major `clap` version upgrade which still has some known unfixed regressions (e.g. `cargo-audit-audit`), and could have other unforeseen changes to how arguments are parsed and usage presented. I think it's worth a minor version bump.

tarcieri (approved):

PR #1108, opened 2024-02-11
Bump rustsec-admin to 0.8.9

amousset:


tarcieri:
Guess we need to bump libgit2 #1109

Shnatsel:
`rustsec-admin` does not depend on libgit2, nor does it activate `rustsec` features that do, so I think that should not be blocking this PR.

The audit job is failing in main too.

Shnatsel (approved):

PR #1110, opened 2024-02-11
Bump rustsec to 0.28.6

amousset:


rustsec/src/advisory.rs
```diff
@@ -63,28 +61,6 @@ impl Advisory {
             .map_err(|e| format_err!(ErrorKind::Parse, "error parsing {}: {}", path.display(), e))
     }
 
-    /// Write an `RUSTSEC-20XX-NNNN.md` file
-    pub fn write_file(&self, path: impl AsRef<Path>) -> Result<(), Error> {
```
amousset, line 67: this was a WIP for full advisory import that will be added in the future.

Shnatsel (approved):

Issue #1107, opened 2024-02-10
Security vulnerability in `libgit2-sys` potentially affecting us through the `fix` feature


Shnatsel:
While most of `cargo audit` is ported to `gitoxide` and is unaffected by [RUSTSEC-2024-0013](https://rustsec.org/advisories/RUSTSEC-2024-0013.html) (thanks, @Byron !), the non-default `fix` feature still pulls in vulnerable versions of `libgit2-sys`:

```
Crate:     libgit2-sys
Version:   0.14.2+1.5.1
Title:     Memory corruption, denial of service, and arbitrary code execution in libgit2
Date:      2024-02-06
ID:        RUSTSEC-2024-0013
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0013
Severity:  8.6 (high)
Solution:  Upgrade to >=0.16.2
Dependency tree:
libgit2-sys 0.14.2+1.5.1
└── git2 0.16.1
    ├── crates-index 0.19.13
    │   └── cargo-edit-9 0.9.3
    │       └── rustsec 0.28.5
    │           ├── rustsec-admin 0.8.8
    │           └── cargo-audit 0.19.0
    └── cargo-edit-9 0.9.3
```

We should probably just rewrite it from the ground up in the long term: https://github.com/rustsec/rustsec/issues/938

It's less clear what to do about it in the short term.

tarcieri:
We could just remove the `fix` feature. I'm not sure it has a whole lot of usage anyway, and it's built on unmaintained dependencies.

PR #1111, opened 2024-02-11
.cargo/audit.toml: ignore RUSTSEC-2024-0013

tarcieri:
It has no immediate resolution due to the `fix` feature using unmaintained/out-of-date dependencies.

We're discussing what to do in #1107 and are definitely aware of the issue. This acknowledges it and allows us to move forward on other development without breaking the build.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,7 +1,6 @@
 [advisories]
 ignore = [
-    "RUSTSEC-2020-0071", # `time` localtime_r segfault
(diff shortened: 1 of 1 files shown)


tarcieri:
This unfortunately didn't actually run the security audit job, but it's failing on `main` anyway so it at least can't make it any worse

Shnatsel:
I am working on a fix in the `rewrite-fix` branch.

Issue #322, opened 2021-03-02
No warning if Cargo.lock is out of date


Nemo157:
```console
> cargo new foo && cd foo
     Created binary (application) `foo` package
> cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 251 security advisories (from /home/wim/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (1 crate dependencies)
> cargo add pest_derive
    Updating 'https://github.com/rust-lang/crates.io-index' index
      Adding pest_derive v2.1.0 to dependencies
> cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 251 security advisories (from /home/wim/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (1 crate dependencies)
```

If I then run a command that forces a lockfile update, you can see that both the count of "crate dependencies" increases, and a vulnerability is found:

```console
> cargo metadata --format-version=1 >/dev/null
    Updating crates.io index
> cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 251 security advisories (from /home/wim/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (21 crate dependencies)
Crate:         generic-array
Version:       0.12.3
Title:         arr! macro erases lifetimes
Date:          2020-04-09
ID:            RUSTSEC-2020-0146
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0146
Solution:      Upgrade to >=0.14.0
Dependency tree:
generic-array 0.12.3
├── digest 0.8.1
│   └── sha-1 0.8.2
│       └── pest_meta 2.1.3
│           └── pest_generator 2.1.3
│               └── pest_derive 2.1.0
│                   └── foo 0.1.0
└── block-buffer 0.7.3
    └── sha-1 0.8.2

error: 1 vulnerability found!
```

IMO `cargo audit` should follow the cargo-subcommand standard of performing a lockfile update itself if not passed `--locked`/`--offline` etc. but at the very least should tell the user if the `Cargo.lock`  is inconsistent with the `Cargo.toml`.

```console
> cargo audit --version
cargo-audit 0.13.1
```

tarcieri:
[...]

PR #1112, opened 2024-02-11
Do not run `cargo update` when auditing

Shnatsel:
`cargo generate-lockfile` behaves like `cargo update` according to its documentation:

> This command will create the Cargo.lock lockfile for the current package or workspace. If the lockfile already exists, it will be rebuilt with the latest available version of every package.

Instead use `cargo update --workspace` that is documented as:

> Attempt to update only packages defined in the workspace. Other packages are updated only if they don’t already exist in the lockfile. This option is useful for updating Cargo.lock after you’ve changed version numbers in Cargo.toml.

I have verified that the `--workspace` option works even there is no workspace.

### DIFF
--- a/cargo-audit/src/lockfile.rs
+++ b/cargo-audit/src/lockfile.rs
@@ -26,7 +26,10 @@ pub fn locate_or_generate(maybe_lockfile_path: Option<&Path>) -> rustsec::Result
 
 /// Run `cargo generate-lockfile`
 pub fn generate() -> rustsec::Result<()> {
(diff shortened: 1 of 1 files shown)


tarcieri:
FWIW `cargo generate-lockfile` was only used in the absence of any `Cargo.lock`. It shouldn't have updated an existing lockfile, or if it did, it would be part of some kind of race condition where two things race to generate `Cargo.lock` since it's absent. But this seems fine anyway.

Shnatsel:
Yeah it's more of a preparation for fixing #322

PR #1113, opened 2024-02-12
Rewrite `cargo audit fix`

### LINKED ISSUE #1107: Security vulnerability in `libgit2-sys` potentially affecting us through the `fix` feature

Shnatsel:
Does away with editing `Cargo.toml`; instead runs `cargo update` behind the scenes to upgrade vulnerable versions.

Fixes  #1107 and  #1109 by ripping out `cargo-edit` which transitively depends on `libgit2-sys` and a bunch of other outdated versions.

### DIFF
--- a/cargo-audit/src/commands/audit/fix.rs
+++ b/cargo-audit/src/commands/audit/fix.rs
@@ -81,6 +81,14 @@ impl Runnable for FixCommand {
                 );
             } else {
                 let mut command = fixer.get_fix_command(vulnerability, dry_run);
+                // If the path to Cargo.lock has been specified explicitly,
+                // run the `cargo update` command in that directory
+                if let Some(path) = self.cargo_lock_path() {
+                    // documentation on .current_dir() recommends canonicalizing the path
+                    let canonical_path = path.canonicalize().unwrap();
+                    let dir = canonical_path.parent().unwrap();
+                    command.current_dir(dir);
+                }
                 // When calling `.status()` the stdout and stderr are inherited from the parent,
                 // so any status or error messages from `cargo update` will automatically be forwarded
                 // to the user of `cargo audit fix`.
--- a/cargo-audit/src/commands/audit/fix.rs
+++ b/cargo-audit/src/commands/audit/fix.rs
@@ -94,13 +94,12 @@ impl Runnable for FixCommand {
                 // to the user of `cargo audit fix`.
                 let status = command.status();
                 if let Err(e) = status {
-                        failed_patches += 1;
-                        status_warn!(
-                            "Failed to run `cargo update` for package {}: {}",
-                            vulnerability.package.name,
-                            e
-                        );
-                    }
+                    failed_patches += 1;
+                    status_warn!(
+                        "Failed to run `cargo update` for package {}: {}",
+                        vulnerability.package.name,
+                        e
+                    );
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


Shnatsel:
Even as `cargo upgrade` is getting integrated into Cargo, changing the version requirements in Cargo.toml in ways other than "just upgrade everything to latest" has been [explicitly deferred](https://blog.rust-lang.org/inside-rust/2024/02/13/this-development-cycle-in-cargo-1-77.html#merging-cargo-upgrade-into-cargo-update).

Therefore I do believe that this approach is the best way forward.

Shnatsel:
@tarcieri I'd appreciate a review from you, since you were the one who authored the previous `fix` implementation.

cargo-audit/src/commands/audit/fix.rs
```diff
@@ -60,34 +57,112 @@ impl Runnable for FixCommand {
...
+                    let dir = canonical_path.parent().unwrap();
+                    command.current_dir(dir);
+                }
+                // When calling `.status()` the stdout and stderr are inherited from the parent,
+                // so any status or error messages from `cargo update` will automatically be forwarded
+                // to the user of `cargo audit fix`.
+                let status = command.status();
+                match status {
+                    Ok(_) => (),
+                    Err(e) => {
```
tarcieri, line 98:
```suggestion
                if let Err(e) = status {
```

rustsec/Cargo.toml
```diff
@@ -39,7 +36,7 @@ serde_json = "1"
 
 [features]
 default = ["git"]
```
tarcieri, line 38: With no additional dependencies, `fix` can probably be enabled by default:

```suggestion
default = ["fix", "git"]
```

Shnatsel: In that case should I remove the feature outright and just make it always enabled?

Shnatsel:
@tarcieri I removed the `fix` feature from `rustsec`. The module is still documented as experimental, and `cargo audit` still doesn't enable its `fix` subcommand by default.

Is this good to go (both to merge and to release so we can get rid of libgit2 CVEs), or is there something else that should be done first?

tarcieri (approved):
Looks good now!

Also FWIW, I didn't write the original: it was contributed in a PR.

PR #1120, opened 2024-02-14
Revert ".cargo/audit.toml: ignore RUSTSEC-2024-0013"

Shnatsel:
Reverts rustsec/rustsec#1111 now that the `git2` dependency is gone for good

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,6 +1,7 @@
 [advisories]
 ignore = [
-    "RUSTSEC-2024-0013", # libgit2-sys RCE, see #1107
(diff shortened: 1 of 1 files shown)


.cargo/audit.toml
```diff
@@ -1,6 +1,7 @@
 [advisories]
 ignore = [
-    "RUSTSEC-2024-0013", # libgit2-sys RCE, see #1107
+    "RUSTSEC-2020-0071", # `time` localtime_r segfault
+    "RUSTSEC-2020-0159", # `chrono` localtime_r segfault
```
tarcieri, line 4: These are no longer necessary either

PR #1143, opened 2024-03-17
Upgrade to gix 0.60 to fix build

Shnatsel:
It is failing for installations from crates.io, so this is a hotfix we need to push ASAP

### DIFF
--- a/rustsec/Cargo.toml
+++ b/rustsec/Cargo.toml
@@ -24,10 +24,10 @@ toml = "0.7"
 url = { version = "2", features = ["serde"] }
 
 # optional dependencies
-tame-index = { version = "0.9.3", default-features = false, features = ["git", "sparse", "native-certs"], optional = true }
+tame-index = { version = "0.9.8", default-features = false, features = ["git", "sparse", "native-certs"], optional = true }
 home = { version = "0.5", optional = true }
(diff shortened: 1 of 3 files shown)


Shnatsel:
Fixing `admin` is more complex so I'm going to leave it as-is for now and focus on `rustsec` and `cargo audit`

rustsec/Cargo.toml
```diff
@@ -24,10 +24,10 @@ toml = "0.7"
 url = { version = "2", features = ["serde"] }
 
 # optional dependencies
-tame-index = { version = "0.9.3", default-features = false, features = ["git", "sparse", "native-certs"], optional = true }
+tame-index = { version = "0.9.8", default-features = false, features = ["git", "sparse", "native-certs"], optional = true }
```
tarcieri, line 27: @Shnatsel the latest twist in this saga is that `tame-index` v0.9.8 and v0.9.9 have been yanked. Looks like we need to bump to v0.10?

Shnatsel: Yeah, I asked for them to be yanked because they are semver-breaking. We do need to bump to v0.10 now which would be gix 0.61 I believe (it's yet another gix bump)

PR #1174, opened 2024-04-16
cargo-audit: remove is-terminal dep; bump gix-* to 0.62

klensy:
cargo-audit msrv is 1.70, IsTerminal stable from 1.70, drop is-terminal dep
updates gix* to 0.62, tame-index to 0.11.

cc https://github.com/rustsec/rustsec/pull/1173

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -290,6 +290,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "base64"
+version = "0.22.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9475866fec1451be56a3c2400fd081ff546538961565ccb5b7142cbd22bc7a51"
+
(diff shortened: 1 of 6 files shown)


tarcieri (approved):
Security audit failure is unrelated but something we should really get fixed

klensy:
gix-transport depend on gix, but
```
# NOTE: Keep in sync with `gix` used by `tame-index`.
gix = { version = "0.61", default-features = false, optional = true }
```
`tame-index` currently uses gix 0.61, waiting on https://github.com/EmbarkStudios/tame-index/pull/58

Shnatsel:
We also need to upgrade our `reqwest` if we're upgrading `gix`, and that may be not entirely trivial.

klensy:
Well, `non_local_definitions` lint was added recently and not exist on stable, so drop in for now.

klensy:
> We also need to upgrade our `reqwest` if we're upgrading `gix`, and that may be not entirely trivial.

Our? this bumps `reqwest` too, but there no direct usage of it, only transitional via `gix`* and `tame-index`.

Shnatsel:
Ah, you are correct! I was mistaken. Thank you!

Issue #1087, opened 2024-01-16
cvss: v4.0 support


tarcieri:
CVSS v4.0 was released on November 1st, 2023:

https://www.first.org/cvss/specification-document#t6

It would be good to add support for it.

amousset:
I'm working on it (and I'll also try to add temporal and environmental groups currently missing from v3).

gildub:
@amousset, can you please provide an update ?

amousset:
I'm back working on this and making good progress, should have something ready this week.

PR #1198, opened 2024-06-06
feat: support temporal score and Environmental Score

bxf12315:


.gitignore
```diff
@@ -1,2 +1,4 @@
 platforms/**/Cargo.lock
 target
+.idea/
```
tarcieri, line 3: Speaking as an IntelliJ user: please edit your `.gitconfig` and add:

```
[core]
  excludesfile = /home/user/.gitignore
```

...and configure your editor-specific errata in there.

Otherwise every repo's `.gitignore` needs to be updated with every file that needs to be ignored by any editor anyone contributes to the project uses.

tarcieri:
It would be nice if these changes could be done in a purely additive manner with any breaking changes as a last resort and kept as minimal as possible.

This crate has been post-1.0 for over 5 years and any unnecessary changes will cause breakages/churn.

bxf12315:
@tarcieri How about this commit?

gildub:
@bxf12315, could you please add a description to this PR ?
I'm trying to help with the review but I got initially confused.
Let's add CVSS v3.1 support for Temporal and Environmental Scores.
And also link related issues ?

gildub (changes requested):
The code looks good to me.
Although the use of abbreviations for the metrics of the various groups is confusing.
It makes sense in a context of a specification document but it's makes the code obscure and hard to read.

For the base group, dubbing the file 'attack_vector.rs', instead of 'av' is clear about its purpose, especially because there are many files such as 'ac', 'av', etc.
But I suppose the main issue is the name of the should reflect the name of the struct or object it manages.

Also I think the use of mod.rs inside of base, environmental and temporal modules would be more idiomatic to Rust as of today.

tarcieri:
Note that `mod.rs` is the older, Rust 2015 convention, and we use the newer Rust 2018 convention with files named after their respective modules: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs

Please don't use `mod.rs`, and if we haven't already, we should enable the lints to deny it.

gildub:
@tarcieri, 

Regarding the mod.rs style, I understand, not worries.

Regarding the file names, shouldn't the file name reflect the struct it contains ? 
That doesn't seems superfluous and not respecting Rust practices.

bxf12315:
> @tarcieri,
> 
> Regarding the mod.rs style, I thought it was the other way around, thanks for the link !
> 
> Regarding the file names, shouldn't the file name reflect the struct it contains ? That doesn't seems superfluous because, for exemple, 'src/v3/temporal.e.rs' file is handling `ExploitCodeMaturity` structure not `e` structure and that doesn't seem to be respecting Rust practices. Also the existing code seems to not have such convention.

@gildub I'm not good at naming, but I personally think that naming conventions consistent with the all project are reasonable.

gildub:
Hi @xiabai,
could you please clarify if you agree with changing the files names to more explicit names or not ?

bxf12315:
> Hi @xiabai, could you please clarify if you agree with changing the files names to more explicit names or not ?

I do not think change file's name is  necessary. Keep changes to a minimum.

> Hi @xiabai, could you please clarify if you agree with changing the files names to more explicit names or not ?

I do not think change file's name is necessary. Keep changes to a minimum.

gildub:
@bxf12315, @tarcieri, 

The argument to use acronyms so it's a minimal change is not valid.
The source code is impacting changes not their file names but here we're talking about extending (adding features) code not bug fixing anyway.

Although names might look just cosmetic there are important especially for long term projects. 
Besides above valid arguments, let's add : 
- Too many files are involved here. We don't want to have to use a mapping table to remember what each file is about. 
- This create a precedent, not a good convention for following versions. 

Look I understand that in Rust every file doesn't necessarily reflect a structure (like it does in Java). Meanwhile here that's the case :

- src/v3/base/a.rs -> Availability
- src/v3/base/a.rs -> `Availability`
- src/v3/base/ac.rs ->  `AttackComplexity`
- src/v3/base/av.rs -> `AttackVector`
- src/v3/base/c.rs -> `Confidentiality`
.../...
- src/v3/environmental/ar.rs -> `AvailabilityRequirement`
- src/v3/environmental/cr.rs -> `ConfidentialityRequirement`

tarcieri:
@gildub the changes you are proposing are, at best, off-topic for this PR. Please open a separate issue.

gildub:
@tarcieri, sorry this is not off topic we're talking about validating this PR. Not doing so will create unnecessary tech debt.

As I mentioned to @xiabai,
I had read the specification document. 
In order to be consistent, if you want to keep the file names with abbreviations, then we need the struct to be name accordingly. For instance you have to replace `Availability` with `A` and so on.

So
`pub enum Availability` becomes pub enum A,
`pub enum AttackComplexitybecomes pub enum AC ,
etc.

So we have to choose we use accronyms all the way or not but we can't mix in both (besides the documentation of course)

bxf12315:
@tarcieri How to trigger the github workflow, I find an compile error, it's just happened on rust 1.6, I want to rebuild it.

tarcieri:
@bxf12315 it should run automatically from now on

gildub:
@tarcieri , I'm trying to help not to slow down the project. I understand that the use of those acronyms might already be a practice, sorry for jumping in from nowhere (with only months of practice in security matters).

I'm not trying to change something already in place, although and again I can't see that in this project.
From years of experience, the inconsistency means tech debt and that will hurt. 
So use acronyms everywhere or don't. That's it. I leave that to you.
And if you think that's relevant beyond this PR then I'm happy to open an issue or not.
Let me know.
Thanks

cvss/src/metric.rs
```diff
@@ -18,7 +20,7 @@ pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
     }
 
     /// Get CVSS v3.1 score for this metric.
-    fn score(self) -> f64;
+    // fn score(self) -> f64;
```
tarcieri, line 23: This is a breaking change

tarcieri:
So yeah, this PR includes many, many breaking changes by decomposing what are currently inherent methods into traits.

That's not necessarily a dealbreaker for this PR but it will signal the start of another breaking release cycle where it would be good to also address #1087, and would therefore delay any usage of these new features outside of prereleases.

I think it can probably still be written in a way that doesn't break any existing APIs (again, this is a v2.0 crate which has been stable for many years), and perhaps introduce those changes separately in a PR that starts a new breaking release cycle.

bxf12315:
> So yeah, this PR includes many, many breaking changes by decomposing what are currently inherent methods into traits.
> 
> That's not necessarily a dealbreaker for this PR but it will signal the start of another breaking release cycle where it would be good to also address #1087, and would therefore delay any usage of these new features outside of prereleases.
> 
> I think it can probably still be written in a purely additive way that doesn't break any existing APIs (again, this is a v2.0 crate which has been stable for many years), and perhaps introduce those changes separately in a PR that starts a new breaking release cycle.



> So yeah, this PR includes many, many breaking changes by decomposing what are currently inherent methods into traits.
> 
> That's not necessarily a dealbreaker for this PR but it will signal the start of another breaking release cycle where it would be good to also address #1087, and would therefore delay any usage of these new features outside of prereleases.
> 
> I think it can probably still be written in a purely additive way that doesn't break any existing APIs (again, this is a v2.0 crate which has been stable for many years), and perhaps introduce those changes separately in a PR that starts a new breaking release cycle.

ok, I got your point, I will drop these three traits MetricScore and ScopedScore, and ModifiedScore, keep the old logic.
I will try to update them before the weekend.

bxf12315:
@tarcieri updated, please check again.

bxf12315:
@tarcieri  please review the PR. Thanks

bxf12315:
@tarcieri hi  please review the PR. Thanks

tarcieri:
@bxf12315 I will try to review soon, however I am the maintainer for a large number of projects and this PR is something of a lower priority to me

cvss/src/metric.rs
```diff
@@ -51,6 +51,48 @@ pub enum MetricType {
 
     /// User Interaction (UI)
     UI,
+
+    /// Exploit Code Maturity (E
```
tarcieri, line 55:
```suggestion
    /// Exploit Code Maturity (E)
```

cvss/src/v3/cvss.rs
```diff
@@ -0,0 +1,212 @@
+use crate::v3::environmental::Environmental;
+use crate::v3::temporal::Temporal;
+use crate::v3::{Base, Score};
+use crate::{Error, MetricType, Severity, PREFIX};
+use core::fmt;
+use core::str::FromStr;
+use std::prelude::rust_2015::{ToOwned, Vec};
```
tarcieri, line 7: Seems oddly auto-imported

tarcieri:
I'm a little confused how referencing `std` didn't break CI

cvss/src/v3/cvss.rs
bxf12315: updated

cvss/src/metric.rs
bxf12315: Done

bxf12315:
Please check again. Thanks

PR #1205, opened 2024-06-24
chore(cargo-lock): bump to toml@0.8

weihanglo:
The only enum having tuple variants is `Error`.
I don't think it is used in serialization/deserialization.

See Changelog:
https://github.com/toml-rs/toml/blob/main/crates/toml/CHANGELOG.md#080---2023-09-13

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,7 +391,7 @@ dependencies = [
  "petgraph",
  "semver",
  "serde",
- "toml 0.7.8",
+ "toml 0.8.10",
  "url",
 ]
 
@@ -3067,6 +3067,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "toml"
+version = "0.8.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9a9aad4a3066010876e8dcf5a8a06e70a558751117a145c6ce2b82c2e2054290"
+dependencies = [
+ "serde",
+ "serde_spanned",
+ "toml_datetime",
+ "toml_edit 0.22.6",
+]
+
+[[package]]
 name = "toml-span"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3104,6 +3116,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c1b5fd4128cc8d3e0cb74d4ed9a9cc7c7284becd4df68f5f940e1ad123606f6"
 dependencies = [
  "indexmap",
+ "serde",
+ "serde_spanned",
  "toml_datetime",
  "winnow 0.6.1",
 ]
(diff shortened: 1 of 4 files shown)


tarcieri:
It's arguably part of `cargo-lock`'s public API by way of a `From<toml::Error>` impl

Shnatsel:
I asked around and it seems the typical way to deal with this is `#[doc(hidden)]`, which I find unsatisfying but is apparently the least invasive way.

We could also restructure it to two types, a private and a public one, and only implement `From` on the public one so we don't have to bump semver every time we bump TOML, and make it the semver guarantees really ironclad. But that's a lot of hassle and boilerplate, and then `?` doesn't quite work anymore, requiring manual `map_err` which I am not keen on either.

tarcieri:
We can just bump the version to 10.0.0 and remove the `From<toml::de::Error>` impl (or rather, change it into a `pub(crate)` function). That should get it out of the public API in perpetuity.

Shnatsel:
I like that approach in principle but I am not keen on writing the `map_err` all over the codebase. I think that's why we didn't go with it for 9.0.0?

But I think that's the least bad option we have. I don't have the time to work on it right now though. @weihanglo would you be willing to contribute that?

tarcieri:
> I think that's why we didn't go with it for 9.0.0?

@Shnatsel I think it was just missed in v9.0.0.

`.map_err(Error:from_toml)` or thereabouts isn't too bad and would keep it out of the public API.

weihanglo:
Done in [`c8b30fd` (#1205)](https://github.com/rustsec/rustsec/pull/1205/commits/c8b30fda18980cc3693cb0f3b10b59719b9fd13a). Is this what we want?

tarcieri:
@weihanglo looks good! Can you bump the version of `cargo-lock` to v0.10.0-pre as well to denote it's a breaking change?

weihanglo:
It seems that  `cargo-lock` and `toml` packages are referenced within the workspace multiple times, and I need to address that. It is easy to get out of sync. Would you want to make use of [the `[workspace.dependencies]` table](https://doc.rust-lang.org/nightly/cargo/reference/workspaces.html#the-dependencies-table) to centralize the dependency information, and after that we can update this PR accordingly?

tarcieri:
Yeah, this is getting to why it hasn't been done yet.

@weihanglo another option is you can temporarily move `cargo-lock` out of the toplevel workspace, by adding a `exclude = ["cargo-lock"]`, so it can have its own separate resolution of `Cargo.lock` which uses a different version of `toml`.

Or, you can bump `toml` everywhere, and bump all the versions to prereleases accordingly (moving it out of the overall workspace would let us do that as a secondary step)

tarcieri:
@weihanglo now that #1209 is landed, can you get this updated? I think you should target an across-the-board breaking `toml` v0.8 update

tarcieri:
Went ahead and did this myself in #1225 and #1226.

@weihanglo thanks for pushing this forward!

weihanglo:
Ah, sorry I have been busy these days. Thanks for picking it up :)

PR #1206, opened 2024-06-24
feat(cargo-lock): support lockfile v4

weihanglo:
In 1.78.0 Cargo introduce v4 lockfile:
https://github.com/rust-lang/cargo/pull/12852

The v3-to-v4 change is minimal: encode URL params with URL encoding.
However, roundtrip test failed because v4 we can only have one
serialization implementation at a time.

I don't know how to proceed. Maybe we could

* Make v4 serialization the default when it becomes the default in Cargo
* Provide API for people to generate different versions of lockfiles.
* Work with upstream Cargo with an in-tree pacakge for lock serialization. (This is on me?)

Opened this PR for discussion.


cargo-lock/tests/examples/Cargo.lock.v4
```diff
@@ -0,0 +1,221 @@
...
+[[package]]
+name = "tinyvec_macros"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cda74da7e1a664f795bb1f8a87ec406fb89a02522cf6e50620d016add6dbbf5c"
+
+[[package]]
+name = "toml"
+version = "0.8.14"
+source = "git+https://github.com/weihanglo/toml.git?branch=a-_%2B%23%24%29z#9e406273177740fa85b86b78e6d5105e932edef0"
```
weihanglo, line 184: Generated from <https://github.com/weihanglo/toml/tree/a-_%2B%23%24)z>.

weihanglo:
Hmm… missing some commits. Wait a sec.

cargo-lock/tests/examples/Cargo.lock.v4
```diff
@@ -0,0 +1,221 @@
...
+[[package]]
+name = "unicode-xid"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8ccb82d61f80a663efe1f787a51b16b5a51e3314d6ac365b08639f52387b33f3"
+
+[[package]]
+name = "url"
+version = "2.5.2"
+source = "git+https://github.com/weihanglo/rust-url.git?tag=a-_%2B%23%24%29z#54346fa288e16b25b71c45149d7067c752b450e0"
```
weihanglo, line 216: Generated from <https://github.com/weihanglo/rust-url/releases/tag/a-_%2B%23%24)z>.

tarcieri:
> Provide API for people to generate different versions of lockfiles.

@weihanglo it already supports this. `Lockfile::version` is consulted at serialization time and used to generate different versions of lockfiles.

weihanglo:
> > Provide API for people to generate different versions of lockfiles.
> 
> @weihanglo it already supports this. `Lockfile::version` is consulted at serialization time and used to generate different versions of lockfiles.

Yeah but not for `SourceId`. We need to make `SourceId` aware of resolve version. On the Cargo side there is an [`EncodableSourceId` doing that job](https://github.com/rust-lang/cargo/pull/12280/files#diff-e85745be0c2cef707bbaa5cbe7cbf80f41b1fdb663c7607f940405c6339930d0R497-R514). I don't think this `cargo-lock` crate is particularly interested of copying over the wrapper from there.

tarcieri:
A lot of the code has been copied from there. It would be good to follow the same approach they're using.

weihanglo:
> Work with upstream Cargo with an in-tree pacakge for lock serialization. (This is on me?)

I am mostly thinking through this. The Cargo team has published a new crate <https://crates.io/crates/cargo-util-schemas> providing basic serialization schemas for common types. I wonder if we could get there so `cargo-lock` can be a thin tool built upon crates from Cargo.

tarcieri:
Sure, that looks potentially useful

tarcieri:
FWIW I took a look at `cargo-util-schemas`. It sure would be nice to use several types defined there, like `PackageIdSpec`, `SourceKind` and `GitReference`, however for those in particular they'd be nice to have as part of the public API, and unfortunately the latter two are enums so they can't really be wrapped in newtypes.

Unfortunately, it doesn't seem ready for that as they just released semver breaking changes a week ago.

In the meantime it doesn't otherwise seem to have types relating specifically to Cargo.lock, only Cargo.toml.

weihanglo:
> In the meantime it doesn't otherwise seem to have types relating specifically to Cargo.lock, only Cargo.toml.

That's true. 

I think maybe we should just copy over from there one more time 😞.

cargo-lock/tests/examples/Cargo.lock.v4
weihanglo: Note that in <3bcf68bca0366154ff957a1a53353716382130c8>, a bug was identified, unrelated to lockfile v4.

weihanglo:
@tarcieri this is ready for review. Take your time :)

Most of the code are copied or adapted from `cargo` the lib.

Note that I've identified a bug not related to lockfile format version: https://github.com/rustsec/rustsec/pull/1206#discussion_r1722172156

tarcieri:
I'm busy for the rest of the week but will try to look this weekend

71:
Note for anyone stumbling on this PR when getting "parse error: invalid Cargo.lock format version: \`4\`" (and who control the workspace whose `Cargo.lock` is causing problems): in the meantime you can add

```toml
rust-version = "1.81" # ensure it stays in lockfile v3
```

to the `Cargo.toml` to prevent Cargo from upgrading the lockfile format to v4 ([source](https://github.com/rust-lang/cargo/blob/5dfdd59009930729787e59049c988a7d18b9f5a1/tests/testsuite/lockfile_compat.rs#L685)).

tarcieri (approved):

tarcieri:
@weihanglo thank you!

PR #1207, opened 2024-06-24
feat(cli): Honor CARGO_TERM_COLOR for cargo-audit

### LINKED ISSUE #865: Respect `CARGO_TERM_COLOR`
I have `CARGO_TERM_COLOR=always` set on CI and I think `cargo audit` should respect it

gilescope:
Adds more 'just works' like any normal cargo command to cargo-audit.
Any explicit command line flag will take priority over `CARGO_TERM_COLOR`.

Fixes #865


Shnatsel:
I like the idea, I'll take a look at the implementation later. Thanks!

PR #1209, opened 2024-07-04
chore: leverage workspace inheritance

weihanglo:
Address <https://github.com/rustsec/rustsec/pull/1205#issuecomment-2189545101>

### DIFF
--- a/.github/workflows/quitters.yml
+++ b/.github/workflows/quitters.yml
@@ -32,7 +32,6 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
           profile: minimal
-      - run: rm ../Cargo.toml # Remove from workspace to make sure `regex` is built without default features
       - run: cargo check
       - run: cargo test --release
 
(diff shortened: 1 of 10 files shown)


weihanglo:
Oops the 1.40.0 MSRV is tricky

Shnatsel:
I think we can just exclude it from the workspace, or bump MSRV if necessary because 1.40 is ancient by now.

weihanglo:
> I think we can just exclude it from the workspace, or bump MSRV if necessary because 1.40 is ancient by now.

7d391035aa7a931ce5e559e17f090e2c94dc6fea fixes it, but I tend to do the latter. Wait a sec.

weihanglo:
@Shnatsel are you good with 1.64? It appears that workspace inheritance was stabilized in 1.64.

https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#workspace-inheritance

weihanglo:
Feel free to merge this as-is if <https://github.com/rustsec/rustsec/commit/7d391035aa7a931ce5e559e17f090e2c94dc6fea> is preferred.

tarcieri (approved):

PR #1227, opened 2024-07-29
cargo-lock v10.0.0-pre.0

tarcieri:


Shnatsel:
Just making sure: you did remove `toml` from `Cargo.lock`'s public API, right? We don't want to go through this all over again next time `toml` bumps semver.

tarcieri:
#1226

Issue #1249, opened 2024-09-29
`cargo-audit`: Support lockfile v4


YBoy-git:
Cargo moves towards using lockfile v4 (https://github.com/rust-lang/cargo/pull/14595).
The current `cargo audit` output for `Cargo.lock` with `version = 4` looks like this:
```
$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 661 security advisories (from /home/yboy/.cargo/advisory-db)
    Updating crates.io index
error: not found: Couldn't load Cargo.lock
Caused by:
  -> I/O operation failed: parse error: parse error: invalid Cargo.lock format version: `4`
```

`cargo` version:
```
$ cargo --version
cargo 1.83.0-nightly (80d82ca22 2024-09-27)
```

`cargo-audit` version:
```
$ cargo audit --version
cargo-audit-audit 0.20.1
```

tarcieri:
See #1206

maxburke:
Not sure if this is the right place to report this but I tried the v0.21.0-rc.0 version of `cargo audit` to try and get the lockfile v4 support but it's panicking:

```
$ RUST_BACKTRACE=1 cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 664 security advisories (from /Users/max/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (851 crate dependencies)
The application panicked (crashed).
Message:  invalid Cargo.lock dependency tree: Resolution("failed to find dependency: datafusion-execution 37.1.0 (git+https://github.com/urbanlogiq/arrow-datafusion?branch=37.1.0-ul)")
Location: /Users/max/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-audit-0.21.0-rc.0/src/presenter.rs:114```

I've attached my lockfile.

[Cargo.lock.gz](https://github.com/user-attachments/files/17440816/Cargo.lock.gz)

**Comment by alex** [comment]
So the error is coming from here: https://github.com/rustsec/rustsec/blob/main/cargo-lock/src/dependency/tree.rs#L45-L47

The source ID for the dependency in question looks like:

```
SourceId {
    url: Url {
        scheme: "https",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Domain(
                "github.com",
            ),
        ),
        port: None,
[...]

PR #1267, opened 2024-10-15
cargo-audit v0.21.0-rc.0

tarcieri:
First prerelease with V4 lockfile support


alex:
Are there any blockers before a final release here?

tarcieri:
Can probably cut one, although it seems the V4 lockfile support may be buggy: #1249

PR #1270, opened 2024-10-24
Attempt to fix an issue with v4 lockfiles

alex:
refs #1249


alex:
This fixes the test case in #1249. But I have no idea if this is correct.

tarcieri:
@alex can you add a test somehow, either writing a unit test or checking in that lockfile and using it for an integration test?

alex:
I added the Cargo.lock file as an integration test. Doing a unit test would probably be better, but I'm remarkably unfamiliar with the cargo-audit/cargo-lock tests, so this was easiest.

PR #1278, opened 2024-10-30
chore: regenerate platform support and bump to platforms@3.4.2

weihanglo:
Ran `regenerate-platforms-crate.sh` for new Rust toolchain changes. Would be great if we can release a new version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2246,7 +2246,7 @@ checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "platforms"
-version = "3.4.1"
+version = "3.4.2"
 dependencies = [
  "serde",
 ]
--- a/platforms/Cargo.toml
+++ b/platforms/Cargo.toml
@@ -4,7 +4,7 @@ description = """
 Rust platform registry with information about valid Rust platforms (target
 triple, target_arch, target_os) sourced from the Rust compiler.
 """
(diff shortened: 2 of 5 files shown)


Shnatsel:
This removes `AARCH64_FUCHSIA` platform so I'm going to bump it to 3.5.0, but otherwise looks good. Thanks!

Shnatsel:
I've published v3.5.0 to crates.io

weihanglo:
Thank you so much!

I feel like maybe we should setup a recurring job in GitHub Action. When a new Rust version got released, it automatically creates a pull request. I only familiar with renovatebot and in Cargo it notifies us when a new Rust release cut: https://github.com/rust-lang/cargo/blob/06e0ef45511b4a773d78379f56f327a5c9524f6e/.github/renovate.json5#L24-L25

(ignore regex, it is only for some wild custom MSRV logic in Cargo's Cargo workspace)

Shnatsel:
Yeah we really should automate it. Rust has a predictable release schedule so it shouldn't be too hard, and we pull data from nightly anyway so even syncing up with the release cycle is not much of a hurdle. We'll also need to detect whether a platform got removed and bump the version based on that, but it should be pretty easy to dump the lists in both versions, run them through `comm` and see if anything got removed. PRs are welcome.

That said, if you need up-to-date target information in an environment where you have rustc available (e.g. in build.rs), you are better off sourcing it from rustc with `rustc --print=target-list` and `rustc --print=cfg --target $MY_TARGET`.

PR #1285, opened 2024-11-09
CVSS v4.0 support

amousset:
Adds support for [CVSS v4.0](https://www.first.org/cvss/v4.0/specification-document).

This is pretty massive (more than I would have liked). The metrics implementation is quite verbose and could probably be reduced with macros, but I'm not convinced it is worth the added complexity, as this code is likely read-only from now.

## API

It reuses little from v3, and introduces new metric types and metric type trait, given the scoring algorithm has changed significantly. The `Score` type is also different, to account for the different rounding and the presence of the `Nomenclature` associated with the value. The only shared parts with the v3 implementation is the `Severity`: it is defined exactly the same, so we can reuse the same type.

It introduces a new `Cvss` type for abstracting over the versions. The common API consists of `FromStr/Display`, `score() -> f64` and `severity()`.

~:grey_question: What do you think about the naming of the new types?. Should we add the version to the type name itself like `ScoreV4`, `MetricsV4`, etc. (in this case we might want to rename the v3 types and deprecate the non-versioned ones), or use the same names behind a `v4::{Score, Vector}` prefix?~

An important change from the v3 is that the v4 API enforces valid vectors, refusing vectors in constructor when it is not possible to score them reliably (i.e., in case of missing, invalid or duplicate metrics). It still allows incorrectly ordered vectors (which are forbidden in the specs) as it does not prevent scoring. The current v3 implementation is more lenient, by allowing to score vectors with missing or duplicate metrics, and uses a default value for the scores when the metrics are not defined, which is not part of the specs.

:grey_question: Should we allow invalid vectors to be stored in `Vector` and make the score method fallible instead?

### Tests

Adding the Red Hat Security repo as submodule to use the test set. Not a fan of submodules but it looks like the most convenient option, and is only required for the tests. Also adding tests for v3 from the test set.

The full test takes around 30 seconds which might be inconvenient, we could discuss restricting the test set is necessary.

## Implementation

It is closely based on [the reference JS implementation](https://github.com/FIRSTdotorg/cvss-v4-calculator/tree/main) (under BSD-2-Clause license).

In order to keep the same kind of strongly-typed implementation, I had to diverge in several places (to put it nicely, the reference implementations rely heavily on the lenient nature of JS and Python). In particular, the metric type API had to be reworked to properly handle the merging of modified values (which actually modified and adds some variants to the enums) and the changes made for computation.

## Compatibility

We add new error variants, but apart from that it should be fully compatible with the existing API. So it can probably be a minor version bump only.

## Next steps

To reach full CVSS coverage, we still have to:

* Add CVSS v2 for completeness (v1 was never really used and is not implemented by other libraries)
* Finish the missing parts of v3 (#1198). We need to converge to a common API around the `Cvss` type.


tarcieri:
> What do you think about the naming of the new types?. Should we add the version to the type name itself like ScoreV4, MetricsV4, etc. (in this case we might want to rename the v3 types and deprecate the non-versioned ones), or use the same names behind a v4::{Score, Vector} prefix?

I think it's generally frowned upon to have the module name in the type name. In cases where you need to deal with them both they can be unambiguously referenced as `v3::Score` and `v4::Score`. See also the [no module prefixes RFC](https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md) which doesn't directly apply here as `ScoreV4` is a suffix, but I think it's a similar idea.

Shnatsel:
Looks like this was good to go for months. Any reason why we still haven't merged this? Anything that needs to happen before then?

tarcieri:
Sorry, this one went through the cracks. I've just wanted to do a pass over the whole thing and review it for consistency with the CVSS v3 implementation, but that's no reason to hold it up for this long

Shnatsel:
I'm not keen on merging a new API without a review, but I also don't want to hold up this functionality for much longer. And I'm not familiar with this code at all.

We already had an advisory DB submission with CVSS 4.0, and I had to delete the CVSS string because 4.0 was not supported by the linter. 

Do you think you could take a look at it sometime next week? If not, we should probably just ship it. I trust Alexis based on his previous PRs.

oxisto:
Is there anything to help support this effort? I am currently trying to integrate CVSS support into https://github.com/csaf-poc/csaf-rust (which parses CSAF documents) and in the long run support for CVSS 4.0 is needed.

I actually would also need support for CVSS 2.0, so I will most likely start a PR about this soon, since the only other rust implementation (https://github.com/moor84/cvssrust) seems to be deserted.

Shnatsel:
At this point we should just Ship It to unblock users, and address any issues as they come up during actual use.

tschmidtb51:
@Shnatsel Thank you so much!

Shnatsel:
I've opened #1368 to ship it

PR #1307, opened 2025-01-12
Cargo.lock: bump dependencies

tarcieri:
Updates the following dependencies:

    $ cargo update
    Updating crates.io index
     Locking 133 packages to latest compatible versions
    Updating allocator-api2 v0.2.18 -> v0.2.21
    Updating anstream v0.6.17 -> v0.6.18
    Updating anstyle v1.0.9 -> v1.0.10
    Updating async-compression v0.4.17 -> v0.4.18
    Updating atom_syndication v0.12.4 -> v0.12.6
    Updating bitflags v2.6.0 -> v2.7.0
    Updating borsh v1.5.1 -> v1.5.3
    Updating bstr v1.10.0 -> v1.11.3
    Updating bytes v1.8.0 -> v1.9.0
    Updating cc v1.1.31 -> v1.2.9
    Updating chrono v0.4.38 -> v0.4.39
    Updating clap v4.5.20 -> v4.5.26
    Updating clap_builder v4.5.20 -> v4.5.26
    Updating clap_derive v4.5.18 -> v4.5.24
    Updating clap_lex v0.7.2 -> v0.7.4
      Adding core-foundation v0.10.0
    Updating cpufeatures v0.2.14 -> v0.2.16
    Updating crossbeam-channel v0.5.13 -> v0.5.14
    Updating crossbeam-deque v0.8.5 -> v0.8.6
    Updating crossbeam-utils v0.8.20 -> v0.8.21
    Updating diligent-date-parser v0.1.4 -> v0.1.5
      Adding displaydoc v0.2.5
    Updating errno v0.3.9 -> v0.3.10
    Updating fastrand v2.1.1 -> v2.3.0
    Updating flate2 v1.0.34 -> v1.0.35
    Updating gix-bitmap v0.2.12 -> v0.2.13
    Updating gix-chunk v0.4.9 -> v0.4.10
    Updating gix-command v0.3.10 -> v0.3.11
    Updating gix-config-value v0.14.9 -> v0.14.10
    Updating gix-date v0.9.1 -> v0.9.3
    Updating gix-path v0.10.12 -> v0.10.13
    Updating gix-prompt v0.8.8 -> v0.8.9
    Updating gix-quote v0.4.13 -> v0.4.14
    Updating gix-sec v0.10.9 -> v0.10.10
    Updating gix-validate v0.9.1 -> v0.9.2
    Updating h2 v0.4.6 -> v0.4.7
    Updating hashbrown v0.15.0 -> v0.15.2
[...]

PR #1308, opened 2025-01-12
Suppress useless warnings

Shnatsel:
Unblocks development.

These are problematic both due to `deny(warnings)` and the sheer amount of times they're emitted, drowning out all the actually useful warnings.

### DIFF
--- a/admin/tests/acceptance.rs
+++ b/admin/tests/acceptance.rs
@@ -1,4 +1,4 @@
-#![deny(rivial_casts, unused_qualifications)]
+#![deny(trivial_casts, unused_qualifications)]
 #![forbid(unsafe_code)]
 
 use abscissa_core::testing::prelude::*;
--- a/admin/tests/acceptance.rs
+++ b/admin/tests/acceptance.rs
@@ -1,4 +1,4 @@
-#![deny(warnings, missing_docs, trivial_casts, unused_qualifications)]
+#![deny(rivial_casts, unused_qualifications)]
 #![forbid(unsafe_code)]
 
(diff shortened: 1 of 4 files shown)


admin/tests/acceptance.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(warnings, missing_docs, trivial_casts, unused_qualifications)]
+#![deny(rivial_casts, unused_qualifications)]
```
tarcieri, line 1:
```suggestion
#![deny(trivial_casts, unused_qualifications)]
```

We can probably also relax these to `warn`, but do `-D warnings` in CI

Shnatsel: Yeah, I agree we should get rid of `deny(warnings)` everywhere and use `RUSTFLAGS='-D warnings'`  on CI. That's orthogonal to this PR though - here I'm just relaxing the rules for rustsec-admin acceptance tests.

Shnatsel:
Well, now that I've merged it, CI started failing. How come did it work on this branch but not after merging?

I'm done with this for today. Ugh.

tarcieri:
It looks like CI never actually passed? We need a dependency upgrade PR like #1307 to clear the self-audit

Shnatsel:
Self-audit never passed but all the tests did. And then the tests failed on main somehow.

tarcieri:
The “happy path” test includes a self-audit

Shnatsel:
:facepalm: 

Okay, that explains it. Thank you.

PR #1309, opened 2025-01-12
Fix self audit with minimal MSRV disruption

Shnatsel:
We cannot upgrade away from problematic `idna` without raising MSRV, but this causes the minimum MSRV disruption by leveraging the MSRV-aware resolver for all other crates in Cargo.lock

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,7 +38,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 1.0.109",
- "synstructure",
+ "synstructure 0.12.6",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


Shnatsel:
I think libm caused an undeclared MSRV bump so we're not really getting real MSRV adherence out of the MSRV-aware resolver.

tarcieri:
I think we should just bump MSRV, and longer-term consider a less restrictive MSRV policy, as it's untenable to have an MSRV policy that's stricter than your dependencies

Shnatsel:
Rust 1.84 shipped the MSRV-aware resolver, so we should be able to have a lower MSRV than dependencies after 1.84

tarcieri:
That would still require bumping MSRV to 1.84

Shnatsel:
Closing in favor of  #1311

PR #1316, opened 2025-01-18
Ship new releases to get features and fixes to users

Shnatsel:


rustsec/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name         = "rustsec"
 description  = "Client library for the RustSec security advisory database"
-version      = "0.30.0"
+version      = "0.30.1"
```
tarcieri, line 4: Per the [current MSRV policy](https://github.com/rustsec/rustsec/tree/main/rustsec#minimum-supported-rust-version) this should be a minor version bump as opposed to a patch bump, i.e. `0.31.0`.

Sidebar: perhaps we should revise the MSRV policy to be less restrictive.

Shnatsel: rustsec MSRV is unchanged, it was already 1.73 before this release. Only the cargo-lock MSRV has changed.

tarcieri (approved):

PR #1317, opened 2025-01-18
Update petgraph to 0.7.1

musicinmybrain:
See https://github.com/petgraph/petgraph/blob/petgraph%40v0.7.1/RELEASES.rst. This is a SemVer-incompatible update only because petgraph’s exposed dependency fixedbitset was updated SemVer-incompatibly, but rustsec’s use of the petgraph APIs is straightforward, and no code changes appear to be necessary:

```
$ rg petgraph::
cargo-lock/src/dependency/tree.rs
185:        use petgraph::visit::EdgeRef;

cargo-lock/src/dependency/graph.rs
3:pub use petgraph::{graph::NodeIndex, EdgeDirection};
8:pub type Graph = petgraph::graph::Graph<Package, Dependency>;

cargo-lock/src/bin/cargo-lock/main.rs
13:use petgraph::graph::NodeIndex;
```

Tested with `cargo test --workspace -- --skip lint_advisory_db`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -808,9 +808,9 @@ dependencies = [
 
 [[package]]
 name = "fixedbitset"
-version = "0.4.2"
+version = "0.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ce7134b9999ecaf8bcd65542e436736ef32ddca1b3e06094cb6ec5755203b80"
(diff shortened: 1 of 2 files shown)


tarcieri:
Note: since `petgraph` is part of `cargo-lock`'s public API, this is unfortunately a SemVer breaking change there as well

djc:
@tarcieri how careful do we have to be about semver-incompatible changes for these crates? Are they widely used as libraries? Should we just move forward with this and bump the cargo-lock version to 11?

PR #1328, opened 2025-02-01
update `gix` to v0.70 and `tame-index` index to 0.17.

Byron:
An update to the latest version with minimal changes to accomplish this.

Particularly the GitModificationTime code could be simplified by using higher-level APIs, but I refrained from this as there may be a good reason not to do it that I am unaware of.

This PR was motivated by https://github.com/GitoxideLabs/gitoxide/discussions/1797 and @decathorpe.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,7 +191,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "serde",
- "syn 2.0.85",
+ "syn 2.0.96",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Byron:
It looks like the reason for Rustsec not updating were rustc-version requirements.
Probably this PR should be closed then, sorry for the noise.

tarcieri:
We can bump MSRV, that's fine

Byron:
Alright!

Causing even more noise 😅 I have opened https://github.com/rustsec/rustsec/pull/1329. Please feel free to push into it to change the MSRV or rebase it as needed.

I thought I could reopen this PR which is the only reason I closed it so quickly, but apparently that isn't possible anymore.

PR #1329, opened 2025-02-01
update `gix` to v0.70 and `tame-index` index to 0.17.

Byron:
An update to the latest version with minimal changes to accomplish this.

Particularly the GitModificationTime code could be simplified by using higher-level APIs, but I refrained from this as there may be a good reason not to do it that I am unaware of.

This PR was motivated by https://github.com/GitoxideLabs/gitoxide/discussions/1797 and @decathorpe.

### DIFF
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -60,7 +60,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.74.0
+          toolchain: 1.81.0
           override: true
           profile: minimal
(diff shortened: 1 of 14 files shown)


tarcieri:
```
the package `tame-index` depends on `gix`, with features: `gix-transport` but `gix` does not have these features.
```

Byron:
I thought this has something to do with CI as locally I can't reproduce it. `cargo test` works fine as well.

tarcieri:
It could just be MSRV related. Can you bump the MSRV in `.github/workflows`?

Shnatsel:
I've fixed all the MSRV issues. Thank you!

decathorpe:
Thank you all!

PR #1351, opened 2025-03-29
Introduce Advisory::is_draft() to unify logic

djc:
Added a commit to update dependencies to deal with the audit failure for [RUSTSEC-2025-0009](https://rustsec.org/advisories/RUSTSEC-2025-0009).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,7 +71,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy",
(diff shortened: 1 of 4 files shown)


tarcieri (approved):

djc:
@tarcieri I don't have permission to merge in this repo. We may also need to publish these for them to be used in the advisory-db CI workflows?

tarcieri:
Yeah, `rustsec-admin` needs to be updated

PR #1352, opened 2025-03-30
Update dependencies, bump version, minor refactoring

djc:
Fixes #1355.

PR #1356, opened 2025-04-08
2024

djc:
Need a rustsec release in order to publish rustsec-admin.


rustsec/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name         = "rustsec"
 description  = "Client library for the RustSec security advisory database"
-version      = "0.30.2"
+version      = "0.30.3"
```
djc, line 4: I'm assuming gix and tame-index are not public dependencies, is that true?

tarcieri: I'm uncertain offhand

djc: Okay, looks like the rustsec public API surface is pretty porous and at the very least includes `fs_err` which was semver-incompatibly bumped in #1352, so bumping this to 0.31.

djc:
@tarcieri got time for another round?

djc:
@tarcieri friendly ping?

djc:
@tarcieri thanks! I don't have crates.io ownership for rustsec or rustsec-admin. Could you invite me? (Or maybe, "rustsec/Working Group" instead?)

djc:
> @tarcieri thanks! I don't have crates.io ownership for rustsec or rustsec-admin. Could you invite me? (Or maybe, "rustsec/Working Group" instead?)

Ping? @alex @Shnatsel?

Shnatsel:
Sure. Another option is to just have advisory-db repo pull a git snapshot, the process with publishing to crates.io has always been so unwieldy that nobody wanted to deal with it, resulting in very infrequent updates; maybe we should switch to that regardless.

djc:
@Shnatsel I think I need access to rustsec as well (and potentially some of its dependencies that are published from this repo).

I'll have a look at taking rustsec-admin directly from Git.

PR #1363, opened 2025-05-02
OSV tweaks

djc:
Some tweaks resulting from the investigation of https://github.com/rustsec/advisory-db/issues/2301.


djc:
@tarcieri ping?

tarcieri (approved):

Shnatsel (approved):

PR #1366, opened 2025-05-21
regenerate platform support for 1.87 and bump to platforms@3.6.0

arielb1:
just ran `./regenerate-platforms-crate.sh` using rustc 1.87

quite a few platforms were removed (`avr-unknown-gnu-atmega328`, `i586-pc-nto-qnx700`, `i586-pc-windows-msvc`, `i686-unknown-redox`, `wasm32-wasi`) so doing a minor bump

cc @djc 

A release would be helpful


arielb1:
cc @weihanglo

platforms/src/target/env.rs
```diff
@@ -35,6 +35,12 @@ pub enum Env {
     /// `nto71`
     Nto71,
 
+    /// `nto71_iosock`
+    Nto71_iosock,
```
tarcieri, line 39: I guess this should be:

```suggestion
    Nto71Iosock,
```

There's some code for doing various enum variant mapping where we either need to special case this, or change it to be the default way that snake case is transformed into camel case.

arielb1:
Added a regex for fixing up the names

arielb1:
Do we need for someone else to merge this?

tarcieri:
I can marge it but it will take me a bit to do a release

djc:
@tarcieri can you give me crates.io privileges?

PR #1374, opened 2025-06-10
Make rustsec library compatible with gix's curl backend

dtolnay:
Currently, rustsec with git support is unusable in a dependency graph that also enables gix's curl feature. For example something as simple as:

```toml
[dependencies]
cargo = "0.88.0"
rustsec = "0.30.2"
```

causes `gix-transport` to fail to compile because rustsec enables gix's reqwest backend (indirectly through tame-index) and cargo enables gix's curl backend.

- https://github.com/EmbarkStudios/tame-index/blob/0.21.0/Cargo.toml#L14-L17
- https://github.com/rust-lang/cargo/blob/0.88.0/Cargo.toml#L51

```console
error[E0428]: the name `Impl` is defined multiple times
   --> $CARGO_HOME/registry/src/index.crates.io-1949cf8c6b5b557f/gix-transport-0.45.0/src/client/blocking_io/http/mod.rs:220:1
    |
217 | pub type Impl = curl::Curl;
    | --------------------------- previous definition of the type `Impl` here
...
220 | pub type Impl = reqwest::Remote;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Impl` redefined here
    |
    = note: `Impl` must be defined only once in the type namespace of this module

error: Cannot set both 'http-client-reqwest' and 'http-client-curl' features as they are mutually exclusive
  --> $CARGO_HOME/registry/src/index.crates.io-1949cf8c6b5b557f/gix-transport-0.45.0/src/client/blocking_io/http/mod.rs:26:1
   |
26 | compile_error!("Cannot set both 'http-client-reqwest' and 'http-client-curl' features as they are mutually exclusive");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This PR makes it possible to use rustsec in a project that prefers curl.

```toml
[dependencies]
cargo = "0.88.0"
rustsec = { default-features = false, features = ["gix-curl"] }
```

### DIFF
--- a/.github/workflows/rustsec.yml
+++ b/.github/workflows/rustsec.yml
@@ -39,7 +39,7 @@ jobs:
       - run: cargo check
       - run: cargo test --no-default-features
       - run: cargo test
-      - run: cargo test --all-features
+      - run: cargo test --features=dependency-tree,osv-export,binary-scanning
 
   doc:
     runs-on: ubuntu-latest
@@ -51,4 +51,4 @@ jobs:
           override: true
           profile: minimal
       - uses: Swatinem/rust-cache@v2
-      - run: cargo doc --all-features
+      - run: cargo doc --features=dependency-tree,osv-export,binary-scanning
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -32,7 +32,11 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all --all-features -- -D warnings
+          args: --workspace --all-features --exclude=rustsec -- -D warnings
+      - uses: actions-rs/cargo@v1
+        with:
+          command: clippy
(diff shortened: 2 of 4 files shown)


tarcieri (approved):
Seems like a technically breaking change because `tame-index` no longer implicitly enables `git`, but otherwise looks fine

dtolnay:
Yes I would benefit from a release. This crate is used in https://github.com/facebookincubator/reindeer, which also depends on `cargo`. I had to disable the `rustsec` dependency in https://github.com/facebookincubator/reindeer/commit/805cfddcceedd787132feacce8e1757894611a48 because release 0.32.2 broke us.

djc:
Release prep in #1375.

PR #1375, opened 2025-06-12
rustsec: prepare 0.30.4 release

djc:
For #1374.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2940,7 +2940,7 @@ dependencies = [
 
 [[package]]
 name = "rustsec"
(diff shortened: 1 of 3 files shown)


tarcieri (approved):
I'm still a little worried this is technically a breaking change, but hopefully not one which will actually cause real-world breakages

djc:
> I'm still a little worried this is technically a breaking change, but hopefully not one which will actually cause real-world breakages

Yeah -- it seems niche enough that it seems probably okay? Otherwise, I'm happy to follow up.

PR #1377, opened 2025-06-16
rustsec: raise an error if the database path does not exist

### LINKED ISSUE #1372: cargo audit does not return an error if db directory (--db) doesn't exists
When running the following commands in a powershell on Windows, I am getting no error at all:

```shell
$ cargo audit --db '/dfgdfgdfg/dfgdfgdfg/test' -n
    Loaded 0 security advisories (from /dfgdfgdfg/dfgdfgdfg/test)
    Scanning Cargo.lock for vulnerabilities (11 crate dependencies)
$ echo $?
True
```

or

```shell
$ cargo audit --db 'C:\dfgdfgdfg\dfgdfgdfg\test' -n
    Loaded 0 security advisories (from C:\dfgdfgdfg\dfgdfgdfg\test)
    Scanning Cargo.lock for vulnerabilities (11 crate dependencies)
$ echo $?
True
```

Expected behavior: cargo audit need to refuse to do anything and return a exit with an error code directly.

rustc/cargo version: 1.86.0
cargo audit version: 0.21.2

-----

Through the processing of the logs, I understand that no database was loaded.
But typically, you would expect such situations to trigger an "exit 1" (or equivalent)
Not doing so is misleading and dangerous: it took me some time to realize that I wasn't actually checking anything because I relied on exit code in my automated check.

djc:
Fixes #1372.

### DIFF
--- a/rustsec/src/database.rs
+++ b/rustsec/src/database.rs
@@ -47,22 +47,26 @@ impl Database {
 
         for collection in Collection::all() {
             let collection_path = path.join(collection.as_str());
+            let collection_entry = match fs::read_dir(&collection_path) {
+                Ok(entries) => entries,
+                // The `Rust` collection is currently not useful to end users
+                Err(_) if collection == &Collection::Rust => continue,
+                Err(err) => return Err(err.into()),
+            };
 
-            if let Ok(collection_entry) = fs::read_dir(&collection_path) {
-                for dir_entry in collection_entry {
-                    let dir_entry = dir_entry?;
-                    if !dir_entry.file_type()?.is_dir() {
+            for dir_entry in collection_entry {
+                let dir_entry = dir_entry?;
+                if !dir_entry.file_type()?.is_dir() {
+                    continue;
+                }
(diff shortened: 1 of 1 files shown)


rustsec/src/database.rs
```diff
@@ -47,22 +47,21 @@ impl Database {
 
         for collection in Collection::all() {
             let collection_path = path.join(collection.as_str());
+            let collection_entry = fs::read_dir(&collection_path)?;
 
-            if let Ok(collection_entry) = fs::read_dir(&collection_path) {
-                for dir_entry in collection_entry {
-                    let dir_entry = dir_entry?;
-                    if !dir_entry.file_type()?.is_dir() {
+            for dir_entry in collection_entry {
```
djc, line 52: Is there some scenario where we should not error if the `Crates` entry exists but the `Rust` entry does not? It seems reasonable to me that conceptually, `Database::open()` should get you both kinds of advisories and anything else is a bit dangerous in potentially failing open.

djc: Updated to avoid raising an error for the `Rust` collection.

PR #1389, opened 2025-07-29
build(deps): upgrade `toml` to v0.9

paolobarbolini:


djc:
Also, would be nice to upgrade toml_edit at the same time.

djc:
See #1405.

PR #1393, opened 2025-08-12
Fix new warning from 1.89, update dependencies

djc:
Left toml dependencies to #1389.


djc:
@tarcieri ping?

alex (approved):
didn't review teh Cargo.lock changes, but teh clippy fix looks good at least.

PR #1403, opened 2025-09-03
Update dependencies

djc:
Replaces #1389. Upgrades past https://rustsec.org/advisories/RUSTSEC-2025-0055.html.


djc:
> TOML parse error?

Yes, [filed](https://github.com/toml-rs/toml/issues/1033). Will cut that from this PR for now.

PR #767, opened 2022-11-26
Fixed inconsistency in encoding lockfiles with one registry

UebelAndre:
closes https://github.com/rustsec/rustsec/issues/687

PR #1406, opened 2025-09-04
[cargo-lock] Improve/fix unambiguous source spec removal

Pagten:
In #767 a change was made to make `cargo-lock` more consistent with `cargo` when dealing with lock files that refer to multiple registries. The PR added code to `EncodableLockfile` generation for removing the source spec from a dependency if the target package of that dependency only has a single registry in the lock file.

Unfortunately, I think the change that was made is not correct:
 * It incorrectly removes source specs from a dependency when the target package of that dependency appears multiple times in the lock file with multiple sources, only one of which is a registry. For instance: a package that is imported once from crates.io and once from a git repo.

* It does not remove source specs from a dependency when multiple versions of the target package of that dependency exist in the lock file but each version is sourced from a different registry. The source spec should be removed from the dependency, since it is not needed for disambiguation (and `cargo` omits it).

This PR addresses these issues by removing the code that was added in #767 and extending the disambiguation logic in `EncodableDependency::v2()` instead. The PR also adds additional example `Cargo.lock` files for testing the behavior. Some existing tests are replaced by these new tests, since they cover the same behavior.

I've also attached a zip file containing some `Cargo.lock` files, together with a test script that allows you to test the behavior of `cargo-lock` on these files. The `Cargo.toml` and Cargo config files are also included so you can verify that `cargo` indeed generates those `Cargo.lock` files (hence they should be considered correct). 

[cargo-lock-experiments.zip](https://github.com/user-attachments/files/22140845/cargo-lock-experiments.zip)


Pagten:
It would be great if @illicitonion and/or @UebelAndre can have a look at this PR, since they were involved with the original one.

Also, I actually ran into this while working with `rules_rust`. Bazel and `cargo` were not agreeing about a `Cargo.lock` file due to this issue.

Pagten:
Thanks for the review @djc. I believe I've addressed your comments.

cargo-lock/tests/lockfile.rs
```diff
@@ -198,155 +198,43 @@ mod tree {
...
-/// Test that a lockfile with ambiguous registries have registries encoded for each package which has multiple
 #[test]
-fn encoding_multi_registry() {
-    let lockfile = cargo_lock::Lockfile::from_str(
-        r#"# This file is automatically @generated by Cargo.
-# It is not intended for manual editing.
-version = 3 
-
-[[package]]
-name = "bytes"
```
djc, line 210: Sorry if I wasn't that clear before: I think all existing test cases should be moved to the new mechanism in the first commit.

cargo-lock/tests/lockfile.rs
```diff
@@ -198,155 +198,43 @@ mod tree {
...
-        lockfile,
-        cargo_lock::Lockfile::from_str(lockfile)
-            .unwrap()
-            .to_string(),
-    );
+/// Test logic related to the disambiguation of dependency specifications when
+/// the lock file contains multiple versions of the same crate and/or multiple
+/// sources for the same crate.
+fn source_disambiguation(test_file_suffix: &str) {
+    let original = std::fs::read_to_string(&format!(
```
djc, line 230: Nit: suggest importing `fs` in this module.

djc (commented):
This is pretty close, some remaining feedback (sorry for the iteration).

cargo-lock/tests/lockfile.rs
Pagten: Hmm, the problem is that the `encoding_multi_registry()` doesn't really fit within the new mechanism. That test takes a lock file _without_ source specs for unambiguous dependency specs, passes it through `cargo-lock` and then expects a lock file _with_ source specs for unambiguous dependency specs. In the new mechanism, the input to and the output from `cargo-lock` need to be the same.

If you want, I can move the input of `encoding_multi_registry()` to a separate file and change the test, but then the commit is not a pure refactoring, it already changes the test. Would you prefer that?

cargo-lock/tests/lockfile.rs
Pagten: Done (in the first commit).

cargo-lock/tests/lockfile.rs
djc: > If you want, I can move the input of `encoding_multi_registry()` to a separate file and change the test, but then the commit is not a pure refactoring, it already changes the test. Would you prefer that?

Yes, I think that would work. I'd like to be able to review in isolation how the old test is being represented in the new situation.

Pagten:
Okay, thanks @djc! Would it be possible to get a new `cargo-lock` release with this change?

djc:
I plan to do a round of releases soon.

cargo-lock/tests/lockfile.rs
tarcieri: Sidebar: I guess all the examples could use `include_str!`

Issue #1384, opened 2025-07-18
feature request: SARIF support to upload to github


gilescope:
Given that github supports uploading security advisories in SARIF format, would be great if we could upload the results into github in that format.

djc:
I'd be happy to review a PR in this direction.

PR #1409, opened 2025-09-08
SARIF API tweaks

djc:
I wanted a bunch of different followup from #1384, so here's what I've got so far.

@cosmir17 @gilescope please have a look?

### DIFF
--- a/cargo-audit/src/lib.rs
+++ b/cargo-audit/src/lib.rs
@@ -28,7 +28,8 @@ pub mod error;
 pub mod lockfile;
 mod prelude;
 pub mod presenter;
-pub mod sarif;
+mod sarif;
+pub use sarif::SarifLog;
 
 /// Current version of the `cargo-audit` crate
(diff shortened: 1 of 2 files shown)


cosmir17 (approved):
I am waiting for @gilescope's review.

djc:
> I've tested this locally and can confirm the SARIF output remains fully compatible with our requirements. All essential fields (tool identification as "cargo-audit", vulnerability details, severity mapping) are preserved and properly formatted.

Thanks for testing!

> One question: With the custom `Serialize` implementations, are we maintaining strict SARIF 2.1.0 schema compliance? Just want to ensure GitHub Security tab and other SARIF consumers will continue to work smoothly.

Do you have any specific concerns? It seems to me that the new implementation should have identical output to the previous one -- it's just a little more efficient.

Shnatsel (approved):
We have a positive review from the consumers of this API and this doesn't touch or affect anything else, so this is safe to merge.

Issue #1323, opened 2025-01-27
[cargo-lock] Usage of `DEFAULT_BRANCH` not consistent with behavior of Cargo


tvsfx:
When parsing lock files, `cargo-lock` sets `master` as the default branch for each git dependency that lacks a branch/revision/tag ([see here](https://github.com/rustsec/rustsec/blob/bd6fb0fba41246ed860e0e8374d8f31aceabc8f9/cargo-lock/src/package/source.rs#L127)). There's two problems with this:

1. This does not correspond to the semantics of cargo git dependencies ([documentation](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)): "Cargo assumes that we intend to use the latest commit on the default branch to build our package if we only specify the repo URL, as in the examples above. " 
Many repositories nowadays use a different default branch than `master`, meaning an incorrect branch will be assumed while parsing. 

2. In my test project, the `source` url in `Cargo.lock` has the following form if one doesn't specify branch/revision/tag for a dependency: `"git+ssh://git@<URL>/<org>/<repo>.git#<revision>"`. In other words, the latest revision from the default branch is included, but no branch name. This is different from specifying the default branch name explicitly, and a correct `Cargo.lock` parser should, imo, distinguish both cases. This would also avoid the (incorrectly) special-cased `Display` logic [here](https://github.com/rustsec/rustsec/blob/bd6fb0fba41246ed860e0e8374d8f31aceabc8f9/cargo-lock/src/package/source.rs#L391).

Would it hence not be more faithful to make the `GitReference::Branch` value optional, or add a constructor for default references to the `GitReference` value? This would, however, constitute a breaking change since `GitReference` has not been marked `#[non_exhaustive]`... A more hacky workaround could be adding a `SourceKind` for default git references, since that enum does have the attribute.

tarcieri:
Regarding the way this question is phrased: there aren't any issues surrounding "faithfulness" since this code is nearly verbatim from an earlier version of `cargo` (see 18a72c1).

Instead upstream `cargo` has diverged, and we have not tracked the changes. Cargo previously hardcoded the `master` branch.

It appears Cargo's handling of default branches changed in this commit: https://github.com/rust-lang/cargo/commit/4c02977c219ef79b5eb09cb1fb39c4068de61559

We can potentially consider re-vendoring the code in a breaking manner since it seems there are other breaking changes we're considering (#1317).

tvsfx:
[...]

PR #1414, opened 2025-09-11
cargo-lock: improve git source normalization

### LINKED ISSUE #1388: Cargo audit crashes when using optional git dependency
#### Summary

When `Cargo.toml` contains **two or more optional dependencies** from the **same git repository**, `cargo-audit` panics during execution. The issue appears to be related to how the `Cargo.lock` dependency tree is resolved.

#### Reproduction Steps

1. Define multiple optional dependencies pointing to the same Git repository (with different `rev`s or crate names).
2. Enable them as part of a feature set.
3. Run `cargo audit`.

#### Example `Cargo.toml` (failing case):

```toml
[features]
tracy = ["tracing-tracy", "tracy-client"]

[dependencies]
tracing-tracy = { rev = "tracing-tracy-v0.10.1", git = "ssh://git@.../tracy_client.git", optional = true }
tracy-client  = { rev = "tracy-client-v0.15.0",  git = "ssh://git@.../tracy_client.git", optional = true }
```

Running `cargo audit` on a project using the `tracy` feature causes a panic:

```
thread 'main' panicked at 'invalid Cargo.lock dependency tree: Resolution("failed to find dependency: tracy-client 0.15.0 (git+ssh://git@.../tracy_client.git?rev=tracy-client-v0.15.0)")'
```

#### Workarounds

* Commenting out either of the two dependencies allows `cargo audit` to run without errors.
* Removing the `optional = true` flag from both dependencies also prevents the panic.
* Using only one of the dependencies in both the `features` and `[dependencies]` sections avoids the issue.

#### Example `Cargo.toml` (working case):

```toml
[features]
tracy = ["tracy-client"]

[dependencies]
tracy-client = { rev = "tracy-client-v0.15.0", git = "ssh://git@.../tracy_client.git", optional = true }
```

#### Expected Behavior

`cargo audit` should correctly process multiple optional git dependencies—even when they originate from the same repository—and not panic.

#### Environment

* `cargo-audit` version: 0.21.2
* `cargo` version: 1.88.0
* OS: Ubuntu 22:02

#### Additional Notes

The issue does not appear to be a problem with Cargo itself, as the project builds and resolves fine with `cargo build`. The crash is specific to `cargo-audit`'s dependency resolution logic.

djc:
Fixes #1388.


tarcieri:
Before cutting another major `cargo-lock` release it’d be good to address #1323 too, since it requires breaking changes that involve vendoring some code from upstream cargo

cargo-lock/src/package/source.rs
```diff
@@ -32,19 +32,53 @@ pub struct SourceId {
     /// The source kind.
     kind: SourceKind,
 
-    /// For example, the exact Git revision of the specified branch for a Git Source.
-    precise: Option<String>,
-
```
tarcieri, line 37: Why was this removed? It still exists upstream: https://github.com/rust-lang/cargo/blob/1ce8023/src/cargo/core/source_id.rs#L56

djc: Basically, I was not aware that the code was copied from Cargo at some point, and from reading the code (and running the tests), it seemed to only be used/relevant in the git case. I could probably revert it. Do you have any other feedback I should process?

tarcieri: The git case may be relevant to other users of `cargo-lock`.

In general that particular file is vendored from cargo (it's noted at the top, although it seems the URL needs to be updated)

djc: It seems to me that the `SourceId` and `SourceKind` stuff in Cargo has gotten a lot more intertwined between 2020 and now, so I'm not sure it's worth doing this. As an alternative, Cargo's cargo-util-schemas crate will include lockfile support in a few releases, so maybe we should drop cargo-lock in favor of using that directly.

tarcieri: Looks like it provides `SourceKind` but not `SourceId`, which doesn't quite seem useful enough to start roping it in, especially since it seems to have breaking changes every few months

PR #1423, opened 2025-09-18
admin: generate per-package OSV JSON files for the site

djc:
Discussed this a while back with the crates.io team.


Shnatsel (approved):
I haven't manually tested it, but LGTM on code level

PR #1428, opened 2025-09-26
cargo-lock: bump version to 11.0.0

djc:
To appease the semver checks that started failing in

- #1425 

Extracted from

- #1414


tarcieri:
@djc hmm, does `cargo-semver-checks` let you bump to a prerelease then so the release bump can happen when you actually tag the release?

PR #1448, opened 2025-10-23
Add new categories: logic-bug and parser-confusion

kpcyrd:
Related to https://github.com/rustsec/advisory-db/pull/2442

Note I'm not a native English speaker, feel free to change/propose better descriptions for those categories.

### DIFF
--- a/rustsec/src/advisory/category.rs
+++ b/rustsec/src/advisory/category.rs
@@ -56,6 +56,11 @@ pub enum Category {
     /// semantically the wrong thing.
     LogicBug,
 
+    /// Parser code is tricked into processing data incorrectly, for example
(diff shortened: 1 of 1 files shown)


tarcieri:
My suggested alternative for rustsec/advisory-db#2442 is `input-validation`, which could be easily retroactively applied to several existing vulnerabilities.

When defining new categories, I'd also suggest looking at high level CWE classes, where `input-validation` maps to [CWE-20](https://cwe.mitre.org/data/definitions/20.html).

tarcieri:
Regarding https://github.com/rustsec/advisory-db/pull/2442 (on second thought here), it reminds me of something like packet-in-packet attacks. I'll have to look into how those are classified in something like CWE.

PR #1456, opened 2025-10-29
cargo-audit: replace most of abscissa with direct calls

djc:


djc:
This should pass the tests. It is still somewhat ugly because all the `status_*!()` macro calls depend on the terminal component being initialized, which in turn requires a live `Application` to initialize.

PR #1465, opened 2025-11-20
feat: support for CVSSv2

### LINKED ISSUE #1379: CVSS 2.0
We are working on a Rust implementation of the CSAF standard (see https://github.com/csaf-poc/csaf-rust). In order to fully comply with the CSAF standard, we need to parse CVSS 2.0, 3.0/3.1 and 4.0 data. Since recently 4.0 was merged in, we are looking forward to use this library and I am considering to contribute with CVSS 2.0. Would that be of interested?

oxisto:
This PR introduces full support for CVSSv2 including temporal and environmental metrics. The style used is somewhat between the v3 and v4 implementation, but leaning more towards v4 where we have a `Vector` struct representing it all and then individual implementations of scores into that vector.

Side node: if interested, I can try to provide the temporal and environmental calculation for v3 as well, in a similar manner.

Fixes #1379

### DIFF
--- a/cvss/Cargo.toml
+++ b/cvss/Cargo.toml
@@ -14,7 +14,8 @@ rust-version = "1.85"
 serde = { workspace = true, optional = true }
 
 [features]
-default = ["std", "v3", "v4"]
+default = ["std", "v2", "v3", "v4"]
+v2 = []
 v3 = []
 v4 = []
 std = []
--- a/cvss/Cargo.toml
+++ b/cvss/Cargo.toml
@@ -1,12 +1,12 @@
 [package]
 name         = "cvss"
 description  = "Common Vulnerability Scoring System parser/serializer"
-version      = "2.2.0"
+version      = "3.0.0"
 license      = "Apache-2.0 OR MIT"
 repository   = "https://github.com/rustsec/rustsec"
 readme       = "README.md"
 categories   = ["parser-implementations"]
-keywords     = ["cvssv3", "cvssv4", "security", "advisory", "vulnerability"]
+keywords     = ["cvssv2", "cvssv3", "cvssv4", "security", "advisory", "vulnerability"]
 edition      = "2024"
 rust-version = "1.85"
 
--- a/cvss/src/v2/metric.rs
+++ b/cvss/src/v2/metric.rs
@@ -1,0 +1,165 @@
+//! CVSS v2.0 metrics.
+
+pub mod base;
+pub mod environmental;
+pub mod temporal;
+
+use alloc::borrow::ToOwned;
+use core::{
+    fmt::{self, Debug, Display},
+    str::FromStr,
+};
+
+use crate::{Error, Result};
+
+/// Trait for CVSS v2.0 metrics.
+pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
+    /// Get the name of this metric.
+    fn name() -> &'static str {
+        Self::TYPE.name()
+    }
+
+    /// Get CVSS v2.0 score for this metric.
+    fn score(self) -> f64;
+
+    /// Get `str` describing this metric's value
+    fn as_str(self) -> &'static str;
+
+    /// [`MetricType`] of this metric.
+    const TYPE: MetricType;
+}
+
+/// Enum over all of the available v3.1 metrics.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[non_exhaustive]
+pub enum MetricType {
+    /// Availability Impact (A)
+    A,
+
+    /// Access Complexity (AC)
+    AC,
+
+    /// Access Vector (AV)
+    AV,
+
+    /// Authentication (Au)
+    Au,
+
+    /// Confidentiality Impact (C)
+    C,
+
+    /// Integrity Impact (I)
+    I,
+
+    /// Exploitability (E)
+    E,
+
+    /// Remediation Level (RL)
+    RL,
+
+    /// Report Confidence (RC)
+    RC,
+
+    /// Collateral Damage Potential (CDP)
+    CDP,
+
+    /// Target Distribution (TD)
+    TD,
+
+    /// Confidentiality Requirement (CR)
+    CR,
+
+    /// Integrity Requirement (IR)
+    IR,
+
+    /// Availability Requirement (AR)   
+    AR,
+}
+
+impl MetricType {
+    /// Get the name of this metric (i.e. acronym)
+    pub fn name(self) -> &'static str {
+        match self {
+            // Base Metrics
+            Self::A => "A",
+            Self::AC => "AC",
+            Self::Au => "Au",
+            Self::AV => "AV",
+            Self::C => "C",
+            Self::I => "I",
+
+            // Temporal Metrics
+            Self::E => "E",
+            Self::RL => "RL",
+            Self::RC => "RC",
+
+            // Environmental Metrics
+            Self::CDP => "CDP",
+            Self::TD => "TD",
+            Self::CR => "CR",
+            Self::IR => "IR",
+            Self::AR => "AR",
+        }
+    }
+
+    /// Get a description of this metric.
+    pub fn description(self) -> &'static str {
+        match self {
+            // Base Metrics
+            Self::A => "Availability Impact",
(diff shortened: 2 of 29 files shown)


djc:
Honestly not sure I'm motivated to review this code on a voluntary basis. Maybe your organization can sponsor this work?

cvss/src/v2/base.rs
```diff
@@ -0,0 +1,67 @@
...
+mod ac;
+mod au;
+mod av;
+mod c;
+mod i;
+
+pub use self::{
+    a::AvailabilityImpact, ac::AccessComplexity, au::Authentication, av::AccessVector,
+    c::ConfidentialityImpact, i::IntegrityImpact,
+};
```
djc: Instead of these `pub use self::` imports, format like this:

```rust
mod a;
pub use a::AvailabilityImpact;

mod ac;
pub use ac::AccessComplexity;
```

etc.

cvss/src/v2/base/a.rs
```diff
@@ -0,0 +1,77 @@
+//! CVSS v2.0 Base Metric Group - Availability Impact (C)
+
+use crate::Error;
+use crate::v2::{Metric, MetricType};
+use alloc::borrow::ToOwned;
+use core::{fmt, str::FromStr};
```
djc: Keep imports in three blocks separated by empty lines.

- `core`, `alloc` and `std`
- External crates
- `crate`

In that order.

cvss/src/v2/metric/base/a.rs
```diff
@@ -0,0 +1,77 @@
...
+    Partial,
+
+    /// Complete (C)
+    /// > There is a total shutdown of the affected resource. The attacker can
+    /// > render the resource completely unavailable.
+    Complete,
+}
+
+impl Metric for AvailabilityImpact {
+    const TYPE: MetricType = MetricType::A;
```
djc: Nit: move associated `const`s to the bottom of the impl block(s).

cvss/tests/redhat-v2.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[test]
+fn cvss_v2_simple() {
+    run_tests_from_file("vectors_simple2");
+}
+
+#[test]
+fn cvss_v2_random() {
+    run_tests_from_file("vectors_random2");
+}
```
djc: Nit: tests before the helper function, please.

djc (commented):
Here's some initial feedback. I'd like this new code to conform to a common style that might not fully align with what's there currently in the crate.

You'll want to rebase on top of `main` and fix any lints (in particular, `clippy::use_self`). `cargo clippy --fix` should be able to take care of most of that.

cvss/src/v2/base.rs
```diff
@@ -0,0 +1,67 @@
...
+use super::Score;
+use crate::v2::{Metric, Vector};
+
+impl Vector {
```
djc: Let's move all the `Vector` methods into a single `impl` block in `vector.rs`, directly below the type definition.

cvss/src/v2/base.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// Calculate Base CVSS score: overall value for determining the severity
+    /// of a vulnerability, generally referred to as the "CVSS score".
+    ///
+    /// Described in CVSS v2.0 Specification: Section 3.2.1:
+    /// <https://www.first.org/cvss/v2/guide#3-2-1-Base-Equation>
+    #[cfg(feature = "std")]
```
djc: Why do we need `std` for these?

djc:
Please squash all the commits into a single one.

cvss/src/v2/base.rs
oxisto: Fine by me, I was actually just following the style in the existing 3.0 files. I can probably adjust this in 3.0 in the other PR as well.

cvss/src/v2/metric/base/a.rs
oxisto: Are you sure about that? That would explicitly not match the structure how it is in v3 and v4?

djc: Yes, I'm sure -- fine with the discrepancy for now.

cvss/src/v2/base.rs
oxisto: Because of `std::f64::round`

djc:
(Add a separate commit with a version bump to fix the CI semver job.)

oxisto:
> Please squash all the commits into a single one.

I addressed all the code review comments and did some additional internal audit, that uncovered a few missing pieces, such as integration into OSV. I have kept those changes on a separate commit yet, might make it easier to review.

I did uncover the difference in scoring between FIRST and NIST. FIRST specifies that all the sub scores such  `BaseScore` and `TemperalScore` are rounded to 1 and uses the rounded version to keep calculating (e.g. the temporal and environmental score). For some reason, the NVD NIST calculator takes the unrounded sub score (e.g. from temporal) in the environmental score and then only applies rounding at the end. This produces slightly different results in environmental and temporal scores.

This library currently implements it according to the FIRST spec.

(cc @tschmidtb51)

cvss/src/v2/metric.rs
```diff
@@ -0,0 +1,164 @@
+//! CVSS v2.0 metrics.
+
+use crate::{Error, Result};
```
djc: Nit: keep the `crate` imports in a separate block.

cvss/src/v2/metric.rs
```diff
@@ -0,0 +1,164 @@
...
+};
+
+pub mod base;
+pub mod environmental;
+pub mod temporal;
+
+/// Trait for CVSS v2.0 metrics.
+pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
+    /// [`MetricType`] of this metric.
+    const TYPE: MetricType;
```
djc: Nit: keep the `const` at the bottom of the block (including in all impls).

cvss/src/v2/vector.rs
```diff
@@ -0,0 +1,499 @@
...
+
+use alloc::{borrow::ToOwned, vec::Vec};
+use core::fmt;
+use core::str::FromStr;
+
+#[cfg(feature = "serde")]
+use {
+    alloc::string::{String, ToString},
+    serde::{Deserialize, Serialize, de, ser},
+};
```
djc: Nit: keep one line per top-level crate. The `alloc` imports should go in the core/alloc/std block, above. (Just duplicate the guard.)

cvss/src/v2/vector.rs
```diff
@@ -0,0 +1,499 @@
...
+                    });
+                }
+
+                Ok((id, value))
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        let components = component_vec.iter();
+        let mut metrics = Self {
+            ..Default::default()
```
djc: Can we just write `Self::default()` here?

cvss/src/v2/vector.rs
```diff
@@ -0,0 +1,499 @@
...
+        };
+
+        fn get_value<T: FromStr<Err = Error>>(
+            metric_type: MetricType,
+            current_val: Option<T>,
+            new_val: &str,
+        ) -> Result<Option<T>> {
+            let parsed: T = new_val.parse()?;
```
djc: Nit: write `T::from_str(new_val)?`, please, without the type annotation.

cvss/Cargo.toml
```diff
@@ -1,20 +1,21 @@
...
 name         = "cvss"
 description  = "Common Vulnerability Scoring System parser/serializer"
-version      = "2.2.0"
+version      = "3.0.0"
 license      = "Apache-2.0 OR MIT"
 repository   = "https://github.com/rustsec/rustsec"
 readme       = "README.md"
 categories   = ["parser-implementations"]
-keywords     = ["cvssv3", "cvssv4", "security", "advisory", "vulnerability"]
+keywords     = ["cvssv2","cvssv3", "cvssv4", "security", "advisory", "vulnerability"]
```
djc: Nit: add a space after the new keyword.

djc (approved):
This mostly looks good to me. Please squash all the commits except the version bump, and address the remaining nits.

djc:
Thanks!

PR #1468, opened 2025-11-24
feat: rework CVSS 3.x and introduce temporal and environmental metrics and scores

oxisto:
Reworks CVSS 3x to the structure that the CVSS 2.0 and 4.0 implementation already follows. It also defines temporal and environmental scores and metrics for CVSS 3.x.

Note: this follows the specification and implementation of FIRST (and not NIST). For some reason some CVSS 3.0 vectors in the red hat data set are different - but I check my scores against the FIRST calculator and they are correct. Therefore I am only taking the base score for CVSS 3.0 from the redhat test set.

### DIFF
--- a/cvss/src/cvss.rs
+++ b/cvss/src/cvss.rs
@@ -33,10 +33,10 @@ pub enum Cvss {
     CvssV20(v2::Vector),
     #[cfg(feature = "v3")]
     /// A CVSS 3.0 base vector
-    CvssV30(v3::Base),
+    CvssV30(v3::Vector),
     #[cfg(feature = "v3")]
     /// A CVSS 3.1 base vector
-    CvssV31(v3::Base),
+    CvssV31(v3::Vector),
     #[cfg(feature = "v4")]
     /// A CVSS 4.0 vector
     CvssV40(v4::Vector),
@@ -53,9 +53,9 @@ impl Cvss {
             #[cfg(feature = "v2")]
             Self::CvssV20(base) => base.score().value(),
             #[cfg(feature = "v3")]
-            Self::CvssV30(base) => base.score().value(),
+            Self::CvssV30(vector) => vector.score().value(),
             #[cfg(feature = "v3")]
-            Self::CvssV31(base) => base.score().value(),
+            Self::CvssV31(vector) => vector.score().value(),
             #[cfg(feature = "v4")]
             Self::CvssV40(vector) => vector.score().value(),
         }
@@ -68,9 +68,9 @@ impl Cvss {
             #[cfg(feature = "v2")]
             Self::CvssV20(base) => base.score().severity(),
             #[cfg(feature = "v3")]
-            Self::CvssV30(base) => base.score().severity(),
+            Self::CvssV30(vector) => vector.score().severity(),
             #[cfg(feature = "v3")]
-            Self::CvssV31(base) => base.score().severity(),
+            Self::CvssV31(vector) => vector.score().severity(),
             #[cfg(feature = "v4")]
             Self::CvssV40(vector) => vector.score().severity(),
         }
@@ -82,9 +82,13 @@ impl Cvss {
             #[cfg(feature = "v2")]
             Self::CvssV20(base) => Box::new(base.metrics().map(|(m, v)| (MetricType::V2(m), v))),
             #[cfg(feature = "v3")]
-            Self::CvssV30(base) => Box::new(base.metrics().map(|(m, v)| (MetricType::V3(m), v))),
+            Self::CvssV30(vector) => {
+                Box::new(vector.metrics().map(|(m, v)| (MetricType::V3(m), v)))
+            }
             #[cfg(feature = "v3")]
-            Self::CvssV31(base) => Box::new(base.metrics().map(|(m, v)| (MetricType::V3(m), v))),
+            Self::CvssV31(vector) => {
+                Box::new(vector.metrics().map(|(m, v)| (MetricType::V3(m), v)))
+            }
             #[cfg(feature = "v4")]
             Self::CvssV40(vector) => {
                 Box::new(vector.metrics().map(|(m, v)| (MetricType::V4(m), v)))
@@ -141,9 +145,9 @@ impl FromStr for Cvss {
 
         match (prefix, major_version, minor_version) {
             #[cfg(feature = "v3")]
-            (PREFIX, "3", "0") => v3::Base::from_str(s).map(Self::CvssV30),
+            (PREFIX, "3", "0") => v3::Vector::from_str(s).map(Self::CvssV30),
             #[cfg(feature = "v3")]
-            (PREFIX, "3", "1") => v3::Base::from_str(s).map(Self::CvssV31),
+            (PREFIX, "3", "1") => v3::Vector::from_str(s).map(Self::CvssV31),
             #[cfg(feature = "v4")]
             (PREFIX, "4", "0") => v4::Vector::from_str(s).map(Self::CvssV40),
             (PREFIX, _, _) => Err(Error::UnsupportedVersion {
@@ -162,9 +166,9 @@ impl fmt::Display for Cvss {
             #[cfg(feature = "v2")]
             Self::CvssV20(base) => write!(f, "{}", base),
             #[cfg(feature = "v3")]
-            Self::CvssV30(base) => write!(f, "{}", base),
+            Self::CvssV30(vector) => write!(f, "{}", vector),
             #[cfg(feature = "v3")]
-            Self::CvssV31(base) => write!(f, "{}", base),
+            Self::CvssV31(vector) => write!(f, "{}", vector),
             #[cfg(feature = "v4")]
             Self::CvssV40(vector) => write!(f, "{}", vector),
         }
--- a/cvss/src/v3/base.rs
+++ b/cvss/src/v3/base.rs
@@ -1,20 +1,10 @@
 //! CVSS v3.1 Base Metric Group
 
-mod a;
-mod ac;
-mod av;
-mod c;
-mod i;
-mod pr;
-mod s;
-mod ui;
-
-pub use self::{
-    a::Availability, ac::AttackComplexity, av::AttackVector, c::Confidentiality, i::Integrity,
-    pr::PrivilegesRequired, s::Scope, ui::UserInteraction,
-};
-
 use super::Score;
+use super::metric::base::{
+    AttackComplexity, AttackVector, Availability, Confidentiality, Integrity, PrivilegesRequired,
+    Scope, UserInteraction,
+};
 use crate::{Error, Metric, MetricType, PREFIX, Result};
 use alloc::{borrow::ToOwned, vec::Vec};
 use core::{fmt, str::FromStr};
--- a/cvss/src/v3/metric.rs
+++ b/cvss/src/v3/metric.rs
@@ -1,5 +1,7 @@
 //! CVSS v3.0/v3.1 metrics.
 
+pub mod base;
+
 use crate::{Error, Result};
 use alloc::borrow::ToOwned;
 use core::{
--- a/cvss/src/v3/metric.rs
+++ b/cvss/src/v3/metric.rs
@@ -1,13 +1,16 @@
 //! CVSS v3.0/v3.1 metrics.
 
 pub mod base;
+pub mod environmental;
+pub mod temporal;
 
-use crate::{Error, Result};
 use alloc::borrow::ToOwned;
 use core::{
     fmt::{self, Debug, Display},
     str::FromStr,
 };
+
+use crate::{Error, Result};
 
 /// Trait for CVSSv3 metrics.
 pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
@@ -53,12 +56,55 @@ pub enum MetricType {
 
     /// User Interaction (UI)
     UI,
+
+    /// Exploit Code Maturity (E)
+    E,
+
+    /// Remediation Level (RL)
+    RL,
+
+    /// Report Confidence (RC)
+    RC,
+
+    /// Confidentiality Requirements (CR)
+    CR,
+
+    /// Integrity Requirements (IR)
+    IR,
+
+    /// Availability Requirements (AR)
+    AR,
+
+    /// Modified Attack Vector (MAV)
+    MAV,
+
+    /// Modified Attack Complexity (MAC)
+    MAC,
+
+    /// Modified Privileges Required (MPR)
+    MPR,
+
+    /// Modified User Interaction (MUI)
+    MUI,
+
+    /// Modified Scope (MS)
+    MS,
+
+    /// Modified Confidentiality (MC)
+    MC,
+
+    /// Modified Integrity (MI)
+    MI,
+
+    /// Modified Availability (MA)
+    MA,
 }
 
 impl MetricType {
     /// Get the name of this metric (i.e. acronym)
     pub fn name(self) -> &'static str {
         match self {
+            // Base metrics
             Self::A => "A",
             Self::AC => "AC",
             Self::AV => "AV",
@@ -67,12 +113,31 @@ impl MetricType {
             Self::PR => "PR",
             Self::S => "S",
             Self::UI => "UI",
+
+            // Temporal metrics
+            Self::E => "E",
+            Self::RL => "RL",
+            Self::RC => "RC",
+
+            // Environmental metrics
+            Self::CR => "CR",
+            Self::IR => "IR",
+            Self::AR => "AR",
+            Self::MAV => "MAV",
+            Self::MAC => "MAC",
+            Self::MPR => "MPR",
+            Self::MUI => "MUI",
+            Self::MS => "MS",
+            Self::MC => "MC",
+            Self::MI => "MI",
+            Self::MA => "MA",
         }
     }
 
     /// Get a description of this metric.
     pub fn description(self) -> &'static str {
         match self {
+            // Base metrics
             Self::A => "Availability Impact",
             Self::AC => "Attack Complexity",
             Self::AV => "Attack Vector",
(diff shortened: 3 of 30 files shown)


cvss/src/v3/base.rs
```diff
@@ -15,74 +15,25 @@ pub use self::{
...
-    pub av: Option<AttackVector>,
-
-    /// Attack Complexity (AC)
-    pub ac: Option<AttackComplexity>,
-
-    /// Privileges Required (PR)
-    pub pr: Option<PrivilegesRequired>,
+/// Calling any of its methods will create a [crate::v3::Vector] internally and
+/// delegate the calls to it.
+pub type Base = Vector;
```
djc: If it's just there for backwards compatibility, mark it as `#[deprecated]`?

cvss/src/v3/base.rs
```diff
@@ -15,74 +15,25 @@ pub use self::{
...
-
-    /// Integrity Impact (I)
-    pub i: Option<Integrity>,
-
-    /// Availability Impact (A)
-    pub a: Option<Availability>,
-}
+impl Vector {
+    /// Alias for `base_score()`. Should not be used anymore.
+    pub fn score(&self) -> Score {
```
djc: `#[deprecated]`?

cvss/src/v3/vector.rs
```diff
@@ -0,0 +1,328 @@
...
+use core::{fmt, str::FromStr};
+
+#[cfg(feature = "serde")]
+use {
+    alloc::string::{String, ToString},
+    serde::{Deserialize, Serialize, de, ser},
+};
+
+/// A CVSS 3.x vector, including Base, Temporal, and Environmental metrics.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
```
djc, line 35: If you're going to move a bunch of code around, please do it in a separate commit. (All the other changes should be squashed into a single commit.)

cvss/src/v3/metric.rs
```diff
@@ -8,7 +8,7 @@ use core::{
 };
 
 /// Trait for CVSSv3 metrics.
-pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
+pub trait Metric: Copy + Clone + Debug + Display + Eq + Ord {
```
djc: That looks like a semver-incompatible change. Why do we need it?

cvss/src/v3/metric.rs
```diff
@@ -24,6 +24,16 @@ pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
     fn as_str(self) -> &'static str;
 }
 
+/// Trait for CVSSv3 modified metrics.
+pub trait ModifiedMetric<M: Metric>: Metric {
```
djc, line 33: Why do we need a trait for this? What is this trying to achieve?

cvss/src/cvss.rs
```diff
@@ -27,10 +27,10 @@ pub const PREFIX: &str = "CVSS";
 pub enum Cvss {
     #[cfg(feature = "v3")]
     /// A CVSS 3.0 base vector
-    CvssV30(v3::Base),
+    CvssV30(v3::Vector),
```
djc, line 36: If we're going to rename this in a bunch of places, that should also be a separate commit (the same commit that introduces the deprecated alias).

cvss/src/cvss.rs
```diff
@@ -48,6 +48,7 @@ impl Cvss {
             Self::CvssV30(base) => base.score().value(),
             #[cfg(feature = "v3")]
             Self::CvssV31(base) => base.score().value(),
+            #[cfg(feature = "v3")]
```
djc, line 57: This looks wrong?

djc (commented):
Here's an initial round of feedback. Please also take the relevant stylistic feedback from https://github.com/rustsec/rustsec/pull/1465 into account here.

oxisto:
How should we proceed with this now that we have a new semver version anyway. The v3 is now inconsistent with the v2 and v4 on two aspects:
- The base vector is called `Base` and the other versions implement a `Vector` containing all base, temporal, etc.
- The metrics of v2 and v4 are in a metric sub-module. The v3 ones are directly in the v3 module.

Any preference on how to proceed to harmonise this?

djc:
> How should we proceed with this now that we have a new semver version anyway. The v3 is now inconsistent with the v2 and v4 on two aspects:
> 
> * The base vector is called `Base` and the other versions implement a `Vector` containing all base, temporal, etc.
> * The metrics of v2 and v4 are in a metric sub-module. The v3 ones are directly in the v3 module.
> 
> Any preference on how to proceed to harmonise this?

Making it consistent sounds good to me. If you move code around, please do it in a separate commit (most of the iterative improvements to added code should be squashed, though).

oxisto:
> > How should we proceed with this now that we have a new semver version anyway. The v3 is now inconsistent with the v2 and v4 on two aspects:
> > 
> > * The base vector is called `Base` and the other versions implement a `Vector` containing all base, temporal, etc.
> > * The metrics of v2 and v4 are in a metric sub-module. The v3 ones are directly in the v3 module.
> > 
> > Any preference on how to proceed to harmonise this?
> 
> Making it consistent sounds good to me. If you move code around, please do it in a separate commit (most of the iterative improvements to added code should be squashed, though).

Sounds good. I hope I can manage that (separating the moving from the improvements)

cvss/src/v3/metric.rs
oxisto: The CVSSv3 modified metrics behave a bit complex:
- If they are specified then their value is used (e.g. modified attack complexity)
- If they are not specified their "base" equivalent (e.g. attack complexity) is used

Therefore, we need something that behaves exactly like base metric but gives it a chance to modify the score, if a value is set. It inherits from the base metric's Metric trait.

djc:
@oxisto let me know when this is ready for review.

oxisto:
> @oxisto let me know when this is ready for review.

Yes, sorry for the small delay. It is ready now! I had to rework the commits a little bit because of the squashing, sorry for the noise.

cvss/src/v3/environmental/ar.rs
```diff
@@ -0,0 +1,82 @@
+//! CVSS v3.1 Environmental Metric Group - Availability Requirements (AR)
+
+use crate::{Error, Metric, MetricType};
```
djc: Nit: split the import blocks everywhere?

cvss/src/v3/environmental/ma.rs
```diff
@@ -0,0 +1,62 @@
...
+
+impl FromStr for ModifiedAvailability {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Error> {
+        if s == "X" {
+            Ok(ModifiedAvailability::NotDefined)
+        } else {
+            Ok(ModifiedAvailability::Modified(s.parse()?))
+        }
```
djc: Nit: write this as a `match`, similar to the other implementations:

```rust
Ok(match s {
     "X" => Self::NotDefined,
     _ => Self::Modified(Availability::from_str(s)?),
})
```

(And same for all the similar implementations.)

cvss/src/v3/base.rs
```diff
@@ -9,80 +9,36 @@ mod pr;
...
-use {
-    alloc::string::{String, ToString},
-    serde::{Deserialize, Serialize, de, ser},
-};
+use crate::{Metric, v3::Vector};
 
 #[cfg(feature = "std")]
 use crate::Severity;
 
-/// CVSS v3.1 Base Metric Group
```
djc, line 31: This reorganization does not belong in this commit.

(I think you had more commits before?)

cvss/src/v3/environmental/ma.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::{Error, Metric, MetricType};
+use core::{fmt, str::FromStr};
+
+/// Modified Availability (MA) - CVSS v3.1 Environmental Metric Group
+///
+/// Described in CVSS v3.1 Specification: Section 4.2:
+/// <https://www.first.org/cvss/v3-1/specification-document#4-2-Modified-Base-Metrics>
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub enum ModifiedAvailability {
```
djc: Can we model all this stuff as a single `Modified<T>` type that wraps the inner metric's type?

cvss/src/v3/metric/base/a.rs
```diff
@@ -1,9 +1,11 @@
 //! Availability Impact (A)
 
-use crate::{Error, Metric, MetricType, Result};
 use alloc::borrow::ToOwned;
 use core::{fmt, str::FromStr};
 
+use crate::v3::{Metric, MetricType};
```
djc, line 6: These changes should be squashed into the previous commit.

djc (commented):
The commit history is not what I had in mind. What I'm looking for is roughly:

- One commit that moves code around into different locations.
- One commit that renames `Base` to `Vector` (or whatever).
- One commit that adds new metrics.

Let me know if you have more questions!

oxisto:
> The commit history is not what I had in mind. What I'm looking for is roughly:
> 
> * One commit that moves code around into different locations.
> * One commit that renames `Base` to `Vector` (or whatever).
> * One commit that adds new metrics.
> 
> Let me know if you have more questions!

I'll try, but not sure if this is really feasible.

djc:
Why not?

cvss/src/v3/metric.rs
```diff
@@ -26,6 +29,23 @@ pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
...
+    /// compute the score.
+    fn modified_score(self, base: Option<M>) -> f64;
+}
+
+/// Trait for CVSSv3 Base metrics that have a corresponding "Modified"
+/// Environmental metric.
+pub trait BaseMetric: Metric {
+    /// [`MetricType`] of the corresponding modified metric.
+    const MODIFIED_TYPE: MetricType;
+}
```
djc, line 47: I think all this could be expressed in a simpler way. What are the requirements here/what are the constraints leading to this design?

cvss/src/v3/score.rs
```diff
@@ -26,13 +26,28 @@ impl Score {
...
-        if score_int % 10000 == 0 {
-            Self((score_int as f64) / 100_000.0)
+    /// Round according to the CVSS v3.0 / v3.1 specification depending on
+    /// the minor version. `minor_version == 0` corresponds to 3.0, and
+    /// `minor_version == 1` corresponds to 3.1.
+    #[cfg(feature = "std")]
+    pub fn roundup_for_version(self, minor_version: usize) -> Self {
+        if minor_version == 0 {
+            // CVSS v3.0: round up to nearest tenth
+            Self((self.0 * 10.0).ceil() / 10.0)
```
djc, line 40: Nit: let's do an early return here to minimize rightward drift in the remainder of this function.

cvss/src/v3/vector.rs
```diff
@@ -70,9 +60,58 @@ pub struct Vector {
...
+
+    /// Modified Availability (MA)
+    pub ma: Option<Modified<Availability>>,
 }
 
 impl Vector {
+    /// Alias for `base_score()`.
+    #[cfg(feature = "std")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
+    pub fn score(&self) -> Score {
```
djc, line 111: Why have this?

cvss/src/v3/vector.rs
```diff
@@ -187,13 +198,175 @@ impl Vector {
...
+            .mui
+            .unwrap_or(Modified::NotDefined)
+            .modified_score(self.ui);
+
+        8.22 * av_score * ac_score * pr_score * ui_score
+    }
+
+    #[cfg(feature = "std")]
+    pub(crate) fn is_modified_scope_changed(&self) -> bool {
+        match self.ms {
```
djc, line 316: Suggest something more like:

```rust
let Some(modified) = self.ms else {
     return self.is_scope_changed();
};

..
```

djc (commented):
Commit history seems okay! Some more feedback on other stuff.

djc:
@oxisto ping? Would be nice to get this merged.

cvss/src/v3/vector.rs
oxisto: Well both v2 and v4 expose the base score as `score` so I added this purely for compatibility. I can remove it if its not needed.

oxisto:
> @oxisto ping? Would be nice to get this merged.

Yes, definitely. I was on holidays. Getting back to it now.

djc:
> > @oxisto ping? Would be nice to get this merged.
> 
> 
> 
> Yes, definitely. I was on holidays. Getting back to it now.

Great. Hope you had a good holiday!

cvss/src/v3/metric.rs
oxisto: I wanted to make it more expressive with the extra trait, but it may not be needed. I will simply it.

oxisto:
> > > @oxisto ping? Would be nice to get this merged.
> > 
> > 
> > Yes, definitely. I was on holidays. Getting back to it now.
> 
> Great. Hope you had a good holiday!

Ok, I addressed the remaining things - as an extra commit for easier review. I can squash it into the last commit after you agree with the changes.

cvss/src/v3/vector.rs
djc: Okay, let's keep it for now.

djc:
Please squash them into the relevant commit(s), I'll review them in context.

oxisto:
> Please squash them into the relevant commit(s), I'll review them in context.

I merged it into the last commit, I hope that's ok. It would be a bit of a pain to include it partially in the commits before :(

cvss/src/v3/metric/environmental/modified.rs
```diff
@@ -0,0 +1,130 @@
...
+    ///
+    /// If the modified metric is `NotDefined`, the base metric value is used to
+    /// compute the score.
+    pub fn modified_score(self, base: Option<T>) -> f64 {
+        match self {
+            Self::Modified(v) => v.score(),
+            Self::NotDefined => base.map(|v| v.score()).unwrap_or(0.0),
+        }
+    }
+}
```
djc: This should be above the blanket impl. For both, please inline the `T` bounds (as in `impl<T: BaseMetric + FromStr<Err = Error>`).

(Should we just have `FromStr<Err = Error>` as a supertrait for `BaseMetric`?)

cvss/src/v3/metric/environmental/modified.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Error> {
+        Ok(match s {
+            "X" => Self::NotDefined,
+            _ => Self::Modified(T::from_str(s)?),
+        })
+    }
+}
+
+impl Modified<PrivilegesRequired> {
```
djc: As another inherent impl, this should be above the trait impls but below the generic inherent impl.

cvss/src/v3/metric.rs
```diff
@@ -26,6 +29,13 @@ pub trait Metric: Copy + Clone + Debug + Display + Eq + FromStr + Ord {
     fn as_str(self) -> &'static str;
 }
 
+/// Trait for CVSSv3 Base metrics that have a corresponding "Modified"
+/// Environmental metric.
+pub trait BaseMetric: Metric {
+    /// [`MetricType`] of the corresponding modified metric.
+    const MODIFIED_TYPE: MetricType;
+}
```
djc: Suggest we move this to the `modified` module, below its impls.

(Should it be sealed?)

cvss/src/v3/base.rs
```diff
@@ -1,47 +1,34 @@
...
-/// > of a successful exploit, and represent the consequence to the
-/// > *thing that suffers the impact*, which we refer to formally as the
-/// > *impacted component*.
-/// >
-/// > While the vulnerable component is typically a software application,
-/// > module, driver, etc. (or possibly a hardware device), the impacted
-/// > component could be a software application, a hardware device or a network
-/// > resource. This potential for measuring the impact of a vulnerability other
-/// > than the vulnerable component, was a key feature introduced with
-/// > CVSS v3.0. This property is captured by the Scope metric.
```
djc, line 54: Why did this get removed?

cvss/src/v3/base.rs
```diff
@@ -151,34 +187,6 @@ impl Vector {
         (1.0 - ((1.0 - c_score) * (1.0 - i_score) * (1.0 - a_score)).abs()).into()
     }
 
-    /// Iterate over all defined Base metrics
-    pub fn metrics(&self) -> impl Iterator<Item = (MetricType, &dyn fmt::Debug)> {
```
djc, line 165: Let's keep this in the same place rather than moving it?

cvss/tests/redhat-v3.rs
```diff
@@ -5,36 +5,94 @@ use std::{fs, str::FromStr};
...
+
+        // Test score calculation.
+        assert!(cvss.base_score().value() >= 0.0);
+        assert!(cvss.base_score().value() <= 10.0);
+        assert!(cvss.temporal_score().value() >= 0.0);
+        assert!(cvss.temporal_score().value() <= 10.0);
+        assert!(cvss.environmental_score().value() >= 0.0);
+        assert!(cvss.environmental_score().value() <= 10.0);
+
+        let base_expected: f64 = base_score.trim().parse::<f64>().unwrap_or_else(|e| {
```
djc: Don't think we need the type annotation here?

cvss/tests/redhat-v3.rs
```diff
@@ -5,36 +5,94 @@ use std::{fs, str::FromStr};
...
+            panic!(
+                "Failed to parse base score '{}' for vector '{}': {:?}",
+                base_score, vector, e
+            )
+        });
+        let diff: f64 = cvss.base_score().value() - base_expected;
```
djc: Nor here, and other uses below.

djc (approved):
Okay, I think this is good enough, thanks!

PR #1478, opened 2025-12-02
Update gix to 0.75 and tame-index to 0.25

blinxen:
This PR makes https://github.com/rustsec/rustsec/pull/1477 and https://github.com/rustsec/rustsec/pull/1475 obsolete. `gix` and `tame-index` need to be updated together anything.


djc (approved):
Thanks!

PR #1480, opened 2025-12-08
Add field expect-deleted to advisory metadata

### LINKED ISSUE #1479: Support for creating advisories for deleted crates
Via the #t-crates-io zulip: https://blog.rust-lang.org/2025/12/03/crates.io-malicious-crates-evm-units-and-uniswap-utils/

Right now the linter checks that crates in advisories actually exist on crates.io, however there are cases like this where the relevant crates have been deleted from crates.io but may still appear in people's `Cargo.lock` where it would still make sense to publish advisories.

I'm thinking we could add some field to the advisory to inform the linter the relevant crate has been deleted, and not to run the normal checks that the crate exists, something like:

```toml
package-deleted = true
```

or

```toml
skip-lint = true
```

djc:
Fixes #1479.

Tested by running the linter against:

- https://github.com/rustsec/advisory-db/pull/2508

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2926,7 +2926,7 @@ dependencies = [
 
 [[package]]
 name = "rustsec"
-version = "0.31.0"
+version = "0.32.0"
 dependencies = [
  "auditable-info",
  "auditable-serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ platforms = { version = "3", path = "./platforms" }
 quitters = { version = "0.1.0", path = "./quitters" }
 regex = { version = "1.10.6", default-features = false }
 rust-embed = "8.5.0"
-rustsec = { version = "0.31", path = "./rustsec" }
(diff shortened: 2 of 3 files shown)


admin/src/linter.rs
```diff
@@ -134,9 +127,11 @@ impl Linter {
...
     /// Perform lints that connect to https://crates.io
     fn crates_io_lints(&mut self, advisory: &rustsec::Advisory) -> Result<(), Error> {
-        if !self.name_is_skipped(advisory.metadata.package.as_str())
-            && !self.name_exists_on_crates_io(advisory.metadata.package.as_str())
-        {
+        if advisory.metadata.expect_deleted {
+            return Ok(());
+        }
+
+        if !self.name_exists_on_crates_io(advisory.metadata.package.as_str()) {
```
riking: This should be `if self.name_exists_on_crates_io(..) == advisory.metadata.expect_deleted`

PR #1529, opened 2026-02-05
admin: don't rely on cached data

djc:
Fixes a regression from:

- #1525

That PR used `cached_krate()` everywhere, which meant that most crates were not populated and yielded lint errors for failing to find a matching package on the index. Fix it by using `krate()` instead -- avoid async code, which is awkward in the context of abscissa.


tarcieri:
> avoid async code, which is awkward in the context of abscissa.

You can always use `abscissa-tokio` if need be

PR #1542, opened 2026-02-18
Show affected functions in binaries

### LINKED ISSUE #1531: Should binary scanning consider symbols?
Advisories have an [`affected_functions` table](https://github.com/rustsec/advisory-db/blob/1c176c03f6d76499a541a4598932837f9f4642ad/EXAMPLE_ADVISORY.md?plain=1#L22-L23). If a binary's symbol table does not list any of the affected functions, should `cargo-audit` not emit a warning for that binary?

I've done some experiments with this, and I'm wondering if a PR would be desired. I pose some questions below, though.

To give a specific example, consider [RUSTSEC-2020-0071: Potential segfault in the time crate](https://rustsec.org/advisories/RUSTSEC-2020-0071) and consider this code:
```toml
# Cargo.toml
[package]
name = "vulnerable-binary"
version = "0.1.0"
edition = "2024"

[dependencies]
time = { version = "~0.1" }
```
```rs
// src/main.rs
fn main() {
    println!("{:?}", time::empty_tm());
    // println!("{:?}", time::now());
}
```
Compiling the binary and running `cargo-audit` on it produces a warning today, even with the call to `time::now` commented out. What I propose is that a warning be emitted only if the call to `time::now` is _not_ commented out.

(I expected [LTO](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) to have to be enabled to witness a difference. But it seems that just commenting out the call to `time::now` is enough to prevent it from appearing in the binary's symbol table.)

---

I've noticed that many existing advisories' affected functions do not exactly match what appear in binaries. To give an example, [RUSTSEC-2026-0009: Denial of Service via Stack Exhaustion](https://rustsec.org/advisories/RUSTSEC-2026-0009.html) lists:
```
time::Date::parse
```
But when a vulnerable version is depended on, the symbol that actually appears in the binary is:
```
time::date::Date::parse
```
Use of parameterized types provides additional examples. The symbols that appear in a binary vulnerable to [RUSTSEC-2025-0016:
Use after free in `Parc`...](https://rustsec.org/advisories/RUSTSEC-2025-0016.html) use `Parc<T>`, not `Parc`.

Arguably, advisories should not be modified to facilitate an enhanced binary-scanning feature. An advisory will make more sense to a user if it names symbols as a user would import them (e.g., `time::Date::parse`) as opposed to how the symbols are defined (e.g., `time::date::Date::parse`).

But that then begs the question: should some sort of fuzzy match be performed between advisory and binary symbols?

A risk is that a user could falsely believe they are not affected by an advisory because of this feature. That is, the user might get the warning with the feature disabled, and not get the warning with the feature enabled. But, in reality, the user is affected and the fuzzy match is just not matching a symbol in the advisory to a symbol in the user's binary.

smoelius:
Resolves #1531

In #1531, it was proposed that affected functions not be used to make pass/fail decisions, but instead be added to the output. I chose to add an "Affected:" line. Here is an example of what the new output looks like:
```
Crate:     branca
Version:   0.9.2
Title:     Unexpected panic when decoding tokens
Date:      2020-11-29
ID:        RUSTSEC-2020-0075
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0075
Severity:  5.5 (medium)
Solution:  Upgrade to >=0.10.0
Affected:  branca::Branca::decode, branca::decode           # <-- Added
Dependency tree:
branca 0.9.2
└── vulnerable-binary-with-affected-functions 0.1.0
```
The above can be reproduced with the following command:
```sh
cargo run --bin cargo-audit -- audit bin cargo-audit/tests/support/binaries/binary-with-affected-functions
```
Note that `binary-with-affected-functions` is large (64M). I tried to follow the example of the existing `cargo audit bin` test binaries, which are checked into the repo directly.

Please also note that this PR was partially vibe-coded with Claude. I take responsibility for everything Claude wrote, though.

Nits are welcome.


smoelius:
I reverted the changes. The two cases are no longer distinguished.

teor2345 (commented):
This looks fine to me, I just had a question about displaying generics. Might be worth a comment in the actual code to clarify the intent.

cargo-audit/src/presenter.rs
```diff
@@ -385,3 +448,121 @@ impl Presenter {
...
+    const EXPECTED_FUNCTION_PATHS: &[(&str, &[&str])] = &[
+        ("RUSTSEC-2019-0036", &["failure::Fail::__private_get_type_id__"]),
+        ("RUSTSEC-2020-0071", &["time::OffsetDateTime::now_local", "time::OffsetDateTime::try_now_local", "time::UtcOffset::current_local_offset", "time::UtcOffset::local_offset_at", "time::UtcOffset::try_current_local_offset", "time::UtcOffset::try_local_offset_at"]),
+        ("RUSTSEC-2020-0075", &["branca::Branca::decode", "branca::decode"]),
+        ("RUSTSEC-2021-0041", &["parse_duration::parse"]),
+        ("RUSTSEC-2022-0004", &["rustc_serialize::json::Json::from_str"]),
+        ("RUSTSEC-2022-0067", &["lzf::compress", "lzf::decompress"]),
+        ("RUSTSEC-2023-0032", &["ntru::types::PrivateKey::export", "ntru::types::PublicKey::export"]),
+        ("RUSTSEC-2023-0054", &["mail_internals::utils::vec_insert_bytes"]),
+        ("RUSTSEC-2024-0018", &["crayon::utils::object_pool::ObjectPool<H,T>::free"]),
```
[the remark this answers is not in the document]
smoelius: > Could I propose that, rather than make this change, the [linter](https://github.com/rustsec/rustsec/blob/main/rustsec/src/advisory/linter.rs#L245) be updated to flag path parameters? I would be happy to submit a follow-up PR if this one is merged.

I may need backtrack on this. I just made the change and ran it, and two advisories were flagged:
- [RUSTSEC-2024-0018](https://rustsec.org/advisories/RUSTSEC-2024-0018.html) (which you called out above)
- [RUSTSEC-2021-0073](https://rustsec.org/advisories/RUSTSEC-2021-0073.html)

For the latter, this is what appears in its [affected functions](https://github.com/rustsec/advisory-db/blob/4ca563022331b3cad6a8baa5fcf923f07ac57761/crates/prost-types/RUSTSEC-2021-0073.md?plain=1#L15):
```toml
functions = { "prost_types::Timestamp::Into<SystemTime>" = ["<= 0.7.0"] }
```

I am hesitant to remove the type parameter from `Into`.

What do you think is the best path forward?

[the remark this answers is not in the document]
teor2345: Yeah that generic does seem important. Maybe the easiest way forward is to accept/preserve all generics. We can do something more complicated later if there's a strong need for it.

How does preserving generics affect comparisons? Does that mean the advisory needs to match the exact names of the generics in the code/binary?

edit: is there a way to preserve generics in the advisory, but not require them to be present if they're missing?

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,149 @@
+use object::{Object, ObjectSymbol};
+use rustsec::advisory::affected::FunctionPath;
+use std::collections::HashSet;
+use syn::TypePath;
```
djc: Nit: isolate the `std` import in a separate block, first, please.

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,149 @@
...
+        .functions
+        .iter()
+        .filter(|(_path, version_reqs)| {
+            version_reqs
+                .iter()
+                .any(|req| req.matches(&warning.package.version))
+        })
+        .map(|(path, _)| path.clone())
+        .collect();
+    filter_by_symbols(function_paths, symbols)
```
djc: Could this yield an `impl Iterator` instead of collecting into `Vec`? It looks like `filter_by_symbols()` could also just be a filter function...

If you do end up needing to `collect()`, please prefer turbofishing (`::<Vec<_>>()`) over type annotations to keep the uninferred type closer to the collection site.

cargo-audit/src/presenter.rs
```diff
@@ -261,11 +321,7 @@ impl Presenter {
     }
 
     /// Print information about the given vulnerability
-    fn print_vulnerability(
-        &mut self,
-        vulnerability: &rustsec::Vulnerability,
-        tree: &dependency::Tree,
-    ) {
+    fn print_vulnerability(&self, vulnerability: &rustsec::Vulnerability) {
```
djc: Would like to see this (and the similar change for `print_warning()`) as a separate commit (or PR if your prefer).

djc (commented):
Thanks, this looks like a good direction! Here's a bunch of feedback that should be fairly straightforward, will have another look after that is addressed.

Please squash all commits except where there is an independent change (see one of the inline comments below).

Would also be good to stick some screenshots of different cases in the PR description.

cargo-audit/src/binary_scanning.rs
smoelius: I run into trouble as follows. The initial two `if`-statements must become:
```rs
    let Some(affected) = &warning.affected else {
        return empty();
    };
    if affected.functions.is_empty() {
        return empty();
    }
```
But then their types do not match the type of the opaque iterator returned by `filter_by_symbols`.

I feel like I run into this problem often. Is there a way around this that I am not seeing?

djc: Yes, that can be a little tricky and I missed that that was an issue here.

I think we can work around it here -- the second if arguably isn't needed, and you could work around the first one with something like `warning.affected.as_deref().unwrap_or_default()` and chain from that maybe?

smoelius: My description of the problem was poor. `unwrap_or_default` works for the `Vulnerability` case. But for the `Warning` case, `affected` is itself a struct ([`Affected`](https://docs.rs/rustsec/latest/rustsec/advisory/affected/struct.Affected.html)). So, `unwrap_or_default` would try to move out of a shared reference.

Would a use of `LazyLock` like the following be acceptable?
```rust
    pub(crate) fn paths_from_warning(
        &self,
        warning: &Warning,
    ) -> impl Iterator<Item = FunctionPath> {
        static EMPTY: LazyLock<Affected> = LazyLock::new(Default::default);
        let affected = warning.affected.as_ref().unwrap_or(&EMPTY);
        ...
```
If you would prefer to wait for the final submission, I would understand.

djc: Ultimately the solution is similar: we need to get a single type. In this case, `Warning::affected` is a `BTreeMap`, whose `iter()` method yields an `collections::btree::map::Iter` which does implement `Default`. So I think you can do something like `warning.affected.as_ref().map(|functions| functions.iter()).unwrap_or_default()`?

smoelius: Never would I have guessed that would work. Thanks!

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,149 @@
...
+
+fn remove_function_path_parameters(ident: &str) -> &str {
+    if let Some(n) = ident.as_bytes().iter().position(|&x| x == b'<') {
+        &ident[..n]
+    } else {
+        ident
+    }
+}
+
+fn is_subsequence(function_path_segs: &[&str], symbol_segs: &[syn::Ident]) -> bool {
```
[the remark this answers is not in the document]
smoelius: I removed the `_segs` abbreviations but I left the `_seg` abbreviations in `is_subsequence`. Specifically, I thought the following changed code looked weird:
```rs
    for function_path in function_path {
```

smoelius:
> Would also be good to stick some screenshots of different cases in the PR description.

Could you describe the types of case you would like to see?

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,128 @@
...
+            file.symbols()
+                .filter_map(|sym| sym.name().ok())
+                .filter(|name| {
+                    crate_names
+                        .iter()
+                        .any(|crate_name| name.contains(crate_name.as_str()))
+                })
+                .map(|name| format!("{:#}", demangle(name)))
+                .filter_map(|name| parse_str::<TypePath>(&name).ok())
+                .map(|type_path| flatten_type_path(&type_path))
```
djc, line 26: Suggest turning this into a single `filter_map()` call.

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,128 @@
...
+    ) -> impl Iterator<Item = FunctionPath> {
+        self.filter(vulnerability.affected_functions().unwrap_or_default())
+    }
+
+    /// Return the affected function paths that appear in the binary's symbol table based on a warning.
+    pub(crate) fn paths_from_warning(
+        &self,
+        warning: &Warning,
+    ) -> impl Iterator<Item = FunctionPath> {
+        let iter = warning
```
djc, line 44: Suggest avoiding the bindings to `iter` and `function_paths` here and trying to fold it all into a single expression (as for the other case, suggest merging `filter()` and `map()` calls where possible).

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+fn flatten_type_path(type_path: &TypePath) -> Vec<Ident> {
+    let mut idents = Vec::new();
+    if let Some(qself) = &type_path.qself
+        && let Type::Path(inner) = &*qself.ty
+    {
+        idents.extend(flatten_type_path(inner));
+    }
```
djc, line 77: Could you rewrite this to rely on iteration instead of recursion?

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    for segment in &type_path.path.segments {
+        idents.push(segment.ident.clone());
+    }
+    idents
+}
+
+fn function_path_matches_symbol(function_path: &FunctionPath, symbol_idents: &[Ident]) -> bool {
```
djc, line 85: Instead of argument names repeating the type, maybe pick something that clarifies the role instead? Like in this case I think `function_path` could be called `affected`.

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,128 @@
...
+    let function_path: Vec<_> = function_path
```
djc, line 86: Nit: turbofishing instead of type annotation.

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,128 @@
...
+        return function_path.is_empty() && symbol_idents.is_empty();
+    }
+    // First segments must match (crate name).
+    if symbol_idents[0] != function_path[0] {
+        return false;
+    }
+    // Last segments must match (function name).
+    if symbol_idents[symbol_idents.len() - 1] != function_path[function_path.len() - 1] {
+        return false;
+    }
```
djc, line 100: It would be nice here to do a `match` expression for the 0, 1, 2, and more cases, instead? Like, I think you can `match (symbol_idents, &function_paths) { ([], []) => false, .. }`.

(Also here, use argument names to clarify the role instead of repeating the type.)

cargo-audit/src/binary_scanning.rs
```diff
@@ -0,0 +1,128 @@
...
+    let symbol_middle = &symbol_idents[1..symbol_idents.len() - 1];
+    is_subsequence(function_path_middle, symbol_middle)
+}
+
+fn remove_function_path_parameters(ident: &str) -> &str {
+    if let Some(n) = ident.as_bytes().iter().position(|&x| x == b'<') {
+        &ident[..n]
+    } else {
+        ident
+    }
```
djc, line 113: Maybe go all in on chaining here with something like `.map(|n| &ident[..n]).unwrap_or(ident)`?

cargo-audit/src/presenter.rs
```diff
@@ -129,16 +147,38 @@ impl Presenter {
             .dependency_tree()
             .expect("invalid Cargo.lock dependency tree");
 
+        #[cfg(feature = "binary-scanning")]
+        let symbols = self.binary_contents.as_deref().and_then(|binary_contents| {
+            let crate_names: HashSet<String> = empty()
```
djc, line 152: Nit: type annotation -> turbofishing (`HashSet<_>` should be sufficient).

djc:
Ignore the self-audit CI failure, will be dealt with in 

- #1580

smoelius:
Please tell me what additional changes you would like.

djc:
Thanks for the iterations -- I'll merge this as is and do some further iterations.

smoelius:
Now that #1588 has been merged, is there anything I can do to help push out a new release of `cargo-audit`?

djc:
Can you submit a PR bumping the version number for cargo-audit? That way I can approve it.

PR #1548, opened 2026-02-24
Hide the link to crates.io for "expect_deleted" crates

amousset:
Don't add a link expected to be broken.


djc:
Are there similar links for the OSV export or in other parts of the generated site (maybe in the RSS feed)?

PR #1560, opened 2026-03-17
Change generated title for malicious crates

LawnGnome:
They're not really "vulnerabilities", so let's be more specific.

This is a very minimal change. I'm open to alternative wording, but I think we should change it _somehow_.

Before:

image

After:

image


amousset (approved):

djc:
(You'll need another admin commit bump to make this go live.)

PR #1577, opened 2026-04-11
fix(cargo-audit): expand tilde in advisory-db path from config

### LINKED ISSUE #585: `cargo audit` creates a `./~/.cargo/advisory-db` directory
Running `cargo audit` on [tendermint-rs](https://github.com/informalsystems/tendermint-rs) locally creates a `~/.cargo/advisory-db` directory within the repository directory itself - it doesn't seem to be resolving the `~` path correctly.

```bash
> cargo audit
# ...

> ls
'~'     Cargo.lock   CHANGELOG.md   CONTRIBUTING.md   img       light-client      light-client-verifier   pbt-gen   README.md    rpc       target       test      TODO
 abci   Cargo.toml   config         docs              LICENSE   light-client-js   p2p                     proto     release.sh   std-ext   tendermint   testgen   tools

> ls ./\~/.cargo/advisory-db
CODE_OF_CONDUCT.md  CONTRIBUTING.md  crates  EXAMPLE_ADVISORY.md  HOWTO_UNMAINTAINED.md  LICENSE.txt  README.md  rust  support.toml
```

Running this on Ubuntu 20.04.4 LTS.

0b0d3:
## Summary

Fixes #585

Shell tilde expansion is not applied to values read from TOML config files. A user who copies `audit.toml.example` and keeps the default `path = "~/.cargo/advisory-db"` ends up with a literal `./~/.cargo/advisory-db` directory created inside their project directory.

## Root cause

The `audit.toml.example` file contained:
\`\`\`toml
path = "~/.cargo/advisory-db"
\`\`\`

When this is deserialized into a `PathBuf`, the `~` is treated as a literal directory name rather than being expanded to the home directory. Shell tilde expansion only happens when the shell parses arguments — not when reading config files.

## Fix

Two-part:

**1. `audit.toml.example`** — comment out the `path` field and document that absolute paths must be used when specifying a custom path. When the field is absent the default (`cargo_home()/advisory-db`) is used, which is always correct.

**2. `auditor.rs`** — add an `expand_tilde()` helper (using the already-present `home` crate) applied to any configured path before use. This makes `~/...` paths work correctly whether set via config file or `--db` flag, and is consistent with user expectations.

## Test plan

- [x] Existing tests updated — `parse_audit_toml_example` now asserts `path` is `None` (commented out in example)
- [x] `cargo test -p cargo-audit` passes
- [x] `cargo build -p cargo-audit` clean with no warnings

### DIFF
--- a/cargo-audit/audit.toml.example
+++ b/cargo-audit/audit.toml.example
@@ -13,7 +13,10 @@ severity_threshold = "low" # CVSS severity ("none", "low", "medium", "high", "cr
 
 # Advisory Database Configuration
 [database]
-path = "~/.cargo/advisory-db" # Path where advisory git repo will be cloned
+# path = "/home/user/.cargo/advisory-db" # Path where advisory git repo will be cloned
+#   Note: use an absolute path here. Shell tilde expansion (~) is not supported
+#   in config files. The default is the `advisory-db` directory inside your
+#   Cargo home (usually $HOME/.cargo/advisory-db).
 url = "https://github.com/RustSec/advisory-db.git" # URL to git repo
 fetch = true # Perform a `git fetch` before auditing (default: true)
 stale = false # Allow stale advisory DB (i.e. no commits for 90 days, default: false)
--- a/cargo-audit/src/auditor.rs
+++ b/cargo-audit/src/auditor.rs
@@ -9,10 +9,32 @@ use rustsec::binary_scanning::BinaryFormat;
 
 use std::{
     io::{self, Read},
-    path::Path,
+    path::{Path, PathBuf},
     process::exit,
     time::Duration,
 };
+
+/// Expand a leading `~/` or bare `~` in a path to the user's home directory.
+///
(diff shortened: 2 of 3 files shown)


cargo-audit/audit.toml.example
```diff
@@ -13,7 +13,10 @@ severity_threshold = "low" # CVSS severity ("none", "low", "medium", "high", "cr
 
 # Advisory Database Configuration
 [database]
-path = "~/.cargo/advisory-db" # Path where advisory git repo will be cloned
+# path = "/home/user/.cargo/advisory-db" # Path where advisory git repo will be cloned
+#   Note: use an absolute path here. Shell tilde expansion (~) is not supported
+#   in config files. The default is the `advisory-db` directory inside your
```
djc, line 18: This is actually fixed in this PR?

Did you self-review your changes before you threw the LLM output over the wall for me to spend time on reviewing?

PR #1583, opened 2026-04-17
Add reachsec as an experimental tool to complement cargo-audit

xizheyin:
As discussed in https://github.com/rustsec/advisory-db/issues/2758#issuecomment-4205492739, I make the tool focus on function reachibility and update the readme.

cc @djc


djc (approved):
This seems okay, thanks!

PR #1588, opened 2026-04-21
Affected binary tweaks

djc:
Followup from

- #1542

Recommend reviewing per-commit.

cc @smoelius

### DIFF
--- a/cargo-audit/src/binary_scanning.rs
+++ b/cargo-audit/src/binary_scanning.rs
@@ -19,18 +19,19 @@ impl SymbolSet {
             .collect::<HashSet<_>>();
 
         let file = File::parse(contents).ok()?;
-        // `parse_str::<TypePath>` is expensive. The filter on `crate_names`
-        // eliminates symbols that we know would be irrelevant.
         Some(Self(
             file.symbols()
                 .filter_map(|sym| {
                     let name = sym.name().ok()?;
+                    // `parse_str::<TypePath>` is expensive. The filter on `crate_names`
+                    // eliminates symbols that we know would be irrelevant.
                     if !crate_names
                         .iter()
                         .any(|crate_name| name.contains(crate_name.as_str()))
                     {
                         return None;
                     }
+
                     let name = format!("{:#}", demangle(name));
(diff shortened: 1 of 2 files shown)


smoelius:
I respect your care for the craft!

djc:
> I respect your care for the craft!

Anything you disagree with/would do differently?

smoelius:
> Anything you disagree with/would do differently?

No, I think I agree with all of them. But thank you very much for asking. 🙏

cargo-audit/src/binary_scanning.rs
```diff
@@ -1,76 +1,81 @@
...
+            let affected = affected
                 .iter()
-                .any(|symbol| function_path_matches_symbol(affected, symbol))
+                .map(|ident| match ident.as_str().split_once('<') {
+                    Some((path, _)) => path,
+                    None => ident.as_str(),
+                })
+                .collect::<Vec<_>>();
+
+            self.0.iter().any(|symbol| {
```
djc, line 61: I experimented with doing a binary search here, but it didn't seem to help at least when running against the test case.

smoelius: Honestly, that surprises me. I just ran a test, and it said the test binary has 10336 symbols. I would expect that to be a level where a difference would be noticeable. I guess other operations dominate the runtime(?). 🤷

djc: I expected the same, but the binary search was a little bit slower. I think making memory access less predictable probably has a decent cost?

smoelius: That sounds plausible.

smoelius:
@djc Do you have a sense for when this will be merged and the feature will be published?

djc:
It's waiting for review. @tarcieri do you have time to take a look? @LawnGnome I'm assuming you're buried in Glasswing review, but would be nice to get this merged.

LawnGnome (approved):
Approved, assuming the failing CI check is going to be fixed when we merge it into the current `main`.

PR #1590, opened 2026-04-22
Take semver-compatible dependency updates

djc:
In particular, rustls-webpki 0.103.13 for https://rustsec.org/advisories/RUSTSEC-2026-0104.html.


djc:
@tarcieri ping?

tarcieri (approved):

PR #1598, opened 2026-05-13
print the invalid char when linting affected functions

Skgland:


djc (approved):
Thanks! Do you want to submit a bump for the rustsec commit used in advisory-db workflows next?

Skgland:
> Thanks! Do you want to submit a bump for the rustsec commit used in advisory-db workflows next?

Opened rustsec/advisory-db#2879

PR #1603, opened 2026-05-31
Validate advisory dates with jiff crate

waamm:
Tackles a 7-year-old TODO by @tarcieri: replace hand-rolled advisory date parsing with ~~`time`~~ `jiff`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3297,6 +3297,7 @@ dependencies = [
  "fs-err",
  "gix",
  "home",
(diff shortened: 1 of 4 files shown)


djc:
Thanks!

I'd prefer to use jiff instead of time.

waamm:
> Thanks!
> 
> I'd prefer to use jiff instead of time.

Ok, done!

djc (approved):

PR #1604, opened 2026-05-31
admin: only rewrite the advisory id field, not lines containing "id = "

### LINKED ISSUE #1602: id assignment broke test code because check for lines containing `id = ` is too simplistic
https://github.com/rustsec/rustsec/blob/a880538b6bf0460c8aa0323ec1ff04c130f87287/admin/src/assigner.rs#L156

that incorrectly applies for lines like `let mut grid = ...`.

broken PR: https://github.com/rustsec/advisory-db/pull/2927/changes
manual fix for broken code: https://github.com/rustsec/advisory-db/pull/2928

StressTestor:
## Problem

The `admin` id assigner rewrites a draft advisory's `id` field with the freshly allocated RustSec id. It did so by rewriting any line matching `current_line.contains("id = ")`:

```rust
if current_line.contains("id = ") {
```

That substring also occurs inside unrelated lines, so example code in the advisory body was clobbered into `id = "RUSTSEC-..."`. This corrupted a real advisory: rustsec/advisory-db#2927 (the line `        let mut grid = ...` was rewritten), manually fixed in rustsec/advisory-db#2928. Fixes #1602.

## Fix

Anchor the check to the start of the line, so only the top-level front-matter `id = ` field is rewritten and indented body lines are left alone. Extracted into a small `is_id_line` helper so the behavior is unit-testable (the surrounding `assign_ids_across_directory` needs a full advisory DB and does filesystem I/O, so it isn't practical to unit-test directly).

Verified `starts_with` is correct against real advisories: the `id` field sits at column 0 (e.g. `rustsec/tests/support/example_advisory_v4*.md`), and the corrupting line in advisory-db#2928 was indented, so the anchored check catches the real field and spares body code.

## Tests

Added unit tests: the real `id = "RUSTSEC-..."` line matches; indented lines that merely contain `id = ` do not. Red/green confirmed (the regression test fails on the old `contains` logic). `cargo clippy --workspace --all-features --exclude=rustsec -- -D warnings`, `cargo fmt --check`, and the `admin` test suite all pass.


djc:
This PR seems to replay all commits or something? Not sure what's going on here, but it doesn't make much sense.

PR #1610, opened 2026-06-05
Bump `cargo-audit` version to 0.22.2

smoelius:
This PR bumps the `cargo-audit` version to 0.22.2 and syncs the lockfile. It makes no other changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,7 +400,7 @@ checksum = "e6e9e01327e6c86e92ec72b1c798d4a94810f147209bbe3ffab6a86954937a6f"
 
 [[package]]
 name = "cargo-audit"
-version = "0.22.1"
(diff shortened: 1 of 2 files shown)


djc:
Why not 0.22.2? Was there some reason to think there are semver-incompatible (that is, non-additive) changes?

smoelius:
> Why not 0.22.2? Was there some reason to think there are semver-incompatible (that is, non-additive) changes?

Apologies. I thought it was still appropriate to bump the minor version for features when the major version is 0.

smoelius:
Thanks, @djc!

djc:
Published [0.22.2](https://github.com/rustsec/rustsec/releases/tag/cargo-audit%2Fv0.22.2).

PR #1613, opened 2026-06-12
platforms: update data

djc:
Comments only, so not planning to release this.


tarcieri:
Not quite sure what this is doing

djc:
> Not quite sure what this is doing

Fair enough. I've updated this PR to switch from a regex-based Markdown parser to use comrak instead, and tweaked the output to get rid of the footnote references, which don't make sense in the context of the platforms crate.

There's a new platform now as well, so bumped the version to prepare for release.

djc:
@tarcieri @LawnGnome friendly ping?

tarcieri (approved):

PR #1619, opened 2026-06-24
cargo-audit: suggest unaffected newer versions

### LINKED ISSUE #1618: `cargo audit` should surface `unaffected` versions when no patched upgrade is available
When `cargo audit` reports an advisory with no patched versions, the human-readable output currently appears to ignore `unaffected` version ranges. This can make the remediation guidance misleading.

Example:

```text
Crate:     hickory-proto
Version:   0.25.2
Title:     NSEC3 closest-encloser proof validation enters unbounded loop on cross-zone responses
Date:      2026-05-01
ID:        RUSTSEC-2026-0118
URL:       https://github.com/hickory-dns/hickory-dns/security/advisories/GHSA-3v94-mw7p-v465
Solution:  No fixed upgrade is available!
```

However, the advisory metadata for `RUSTSEC-2026-0118` includes unaffected ranges for `hickory-proto`:

```text
unaffected:
  <0.25.0-alpha.3
  >=0.26.0-beta.1
```

The advisory text explains that the affected code was migrated out of `hickory-proto` into `hickory-net`. So while there may be no “patched” `hickory-proto` version in the affected line, there is still an actionable migration path: move off the affected `0.25.x` line to an unaffected `hickory-proto` version, and separately audit/use the relevant `hickory-net` version if the migrated functionality is needed.

The current output:

```text
Solution: No fixed upgrade is available!
```

is technically understandable if it only means “no patched version exists”, but from a user/operator perspective it strongly suggests “there is no upgrade path and nothing actionable to do.” In my case, this caused me to defer remediation because I assumed I had to wait for a future fixed release that would never arrive.

Suggested behavior:

When an advisory has no patched versions but does have unaffected ranges, `cargo audit` should show those unaffected ranges in the human-readable output.

For example:

```text
Solution: No patched upgrade is available for this crate line.
Unaffected versions: <0.25.0-alpha.3, >=0.26.0-beta.1
Note: unaffected versions may represent removed or migrated vulnerable code rather than an in-place patch.
```

Or, more compactly:

```text
Solution: No patched upgrade is available.
Unaffected versions exist: <0.25.0-alpha.3, >=0.26.0-beta.1
```

This would make a significant practical difference for advisories where remediation is “migrate to an unaffected version/API” rather than “upgrade to a patched version.”

Expected behavior:

`cargo audit`’s human-readable output should distinguish between:

* no patched version exists;
* no safe/unaffected version exists;
* unaffected versions exist but are not classified as patched.

Actual behavior:

The output only says:

```text
Solution: No fixed upgrade is available!
```

which hides the actionable `unaffected` metadata.

djc:
Fixes #1618.


cargo-audit/tests/acceptance.rs
```diff
@@ -308,6 +314,23 @@ fn notice_advisories_found_json() {
     assert_eq!(advisory_id, "RUSTSEC-2022-0058");
 }
 
+#[test]
+fn unaffected_newer_suggests_upgrades() {
```
LawnGnome, line 318: I have a couple of concerns on this test:

1. I'm not sure it's sufficiently robust against future advisories that might affect `hickory-proto 0.25.1`, since it explicitly checks for 2 upgrade mentions.
2. It doesn't actually test the presentation change in the PR.

Would you consider something like 249ccd99c1aa57b16daff37abee0c85f051084b7 as an expansion of this test? I don't completely love that either, since it invokes `cargo audit` twice, but I didn't want to go too far into the weeds of parsing the human-readable output.

LawnGnome (commented):
The actual implementation LGTM, but I have a question around the acceptance test.

cargo-audit/tests/acceptance.rs
djc: Great idea, thanks for the review! Squashed into my commit with light style editing, added you as a commit co-author.

LawnGnome (approved):

PR #1624, opened 2026-07-01
rustsec: add alternatives field to advisory metadata

djc:
I've been wondering if using this on crates.io in the unmaintained banner would be a nicer experience.

@LawnGnome @Turbo87 thoughts?

Many unmaintained advisories already have a list of recommended alternatives with links to crates.io, so this does not provide a new capability by any means. But it feels like it could make the UX nicer (especially not having to bounce from an unmaintained crate on crates.io -> advisory page -> alternative on crates.io), and makes the alternatives more legible to other code.

### DIFF
--- a/rustsec/src/advisory/metadata.rs
+++ b/rustsec/src/advisory/metadata.rs
@@ -95,4 +95,10 @@ pub struct Metadata {
     /// case, for example if a malicious crate has been completely removed.
     #[serde(rename = "expect-deleted", default)]
     pub expect_deleted: bool,
+
+    /// Recommended alternative crates to use instead of the affected crate.
+    ///
+    /// Typically used with `informational = "unmaintained"` advisories.
+    #[serde(default)]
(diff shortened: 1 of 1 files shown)


rustsec/src/advisory/metadata.rs
```diff
@@ -95,4 +95,10 @@ pub struct Metadata {
     /// case, for example if a malicious crate has been completely removed.
     #[serde(rename = "expect-deleted", default)]
     pub expect_deleted: bool,
+
+    /// Recommended alternative crates to use instead of the affected crate.
+    ///
+    /// Typically used with `informational = "unsound"` advisories.
```
Turbo87: the PR description is talking about "unmaintained" advisories, but the comment says "unsound". is that intentional?

djc: Nope, just my head doing silly stuff.

PR #1625, opened 2026-07-02
Set user agent for index clients

djc:
Since

- #1621 

the GitHub Actions jobs linting advisories have been getting 503 responses. Set user agent values.

@Turbo87 do you think we could get the rate limits for rustsec-admin bumped?


djc:
Upstream fix for the audit findings:

- https://github.com/rust-syndication/atom/pull/95

Turbo87:
> do you think we could get the rate limits for rustsec-admin bumped?

we have burst capacity, so I don't think the rate limit should be much of a problem once the user-agent issue is solved

LawnGnome (approved):

LawnGnome:
Going to merge, since the CI is failing on today's `quick-xml` advisory and I'm sure dependabot will be along shortly.

PR #1627, opened 2026-07-03
Platforms 4

djc:
There were some semver-incompatible changes in the generated platforms data, so I took the opportunity to clean the crate up a little bit. rustsec has so far re-exported the platforms crate, which feels like a semver hazard, so I've removed the re-export for now (since rustsec would have needed a semver-incompatible bump anyway).

https://github.com/rustsec/rustsec/actions/runs/28643503065/job/84944711890


LawnGnome (approved):
I would be interested in the takes of others who have been working in this codebase longer here, just because I feel like there's a bunch of history I'm likely unaware of in terms of how these crates are arranged and the specific `platforms` re-export, but on the face of it I don't have any concerns here.

There's a fair bit of noise, but it's all pretty straightforward reviewing it commit-by-commit, and the non-semver version bumps should take care of downstream users.

So, approved, but I wouldn't mind a quick stamp from @tarcieri and/or @Shnatsel here just in case there's something I'm unaware of. (That said, I'm also fine not blocking on that forever.)

Shnatsel:
I don't understand the motivation for 1704f8b and 38cccd0; but I don't remember why I added those in the first place so I can't really comment. 

My only regret wrt `platforms` is that `rustsec` errors out if it finds unknown platforms in the list of affected platforms in the advisory TOML. It should just skip the unknown entries instead. That error behavior hamstrung our ability to list platforms because some old version that didn't know about some obscure platform yet would error out on loading the database because of it.

djc:
> I don't understand the motivation for [1704f8b](https://github.com/rustsec/rustsec/commit/1704f8b58033661b282b6fb8dc1867f79f50dbaa) and [38cccd0](https://github.com/rustsec/rustsec/commit/38cccd02c11b05e2613a09aa90463506908bcfa8); but I don't remember why I added those in the first place so I can't really comment.

The ignoring of `target` confused me because it also matches `/platforms/src/target`, which seems unintentional. Most Rust repos these days use `/target` (I think Cargo also uses this on `cargo init`). This was added in the initial commit (4ba98a63050bb) so its more general matching seems unintentional to me.

The removal of `platforms/**/Cargo.lock` comes from

- #533 

which nested the platforms-data-gen and platforms-table-gen crates inside the platforms crate. However, in

- #1367

I moved those crates back into the top-level workspace directly so that they can benefit from sharing in workspace dependencies. IMO having all the crates in a flat workspace is more common and therefore more intuitive.

> My only regret wrt `platforms` is that `rustsec` errors out if it finds unknown platforms in the list of affected platforms in the advisory TOML. It should just skip the unknown entries instead. That error behavior hamstrung our ability to list platforms because some old version that didn't know about some obscure platform yet would error out on loading the database because of it.

I'll review this after this.

Shnatsel:
The gitignore changes sound good, thank you for the explanation.

tarcieri:
Chiming in here a bit too late to say I agree with @Shnatsel that it would be nice if more fields in the advisories were `String` instead of a specific schema so the latter can evolve with time, and the linter were used instead to ensure it does fit the schema. Then you could completely remove `platforms` from the public API of the `rustsec` crate.

PR #1630, opened 2026-07-08
rustsec: lint against bare usernames in advisory Markdown

djc:
Fixes

- https://github.com/rustsec/advisory-db/issues/3046

@taiki-e sorry for the notification flood. Do you think the last commit here would be sufficient?

Have you tested that the explicit links don't trigger notifications?

### DIFF
--- a/rustsec/src/advisory/linter.rs
+++ b/rustsec/src/advisory/linter.rs
@@ -64,6 +64,50 @@ impl Linter {
             errors: vec![],
         };
 
+        // Lint against user mentions to avoid spamming
+        if advisory_parts.markdown.contains("@") {
+            let front_matter_lines = advisory_parts.front_matter.lines().count();
+            for (i, line) in advisory_parts.markdown.lines().enumerate() {
+                let mut found = None;
+                for (pi, part) in line.split_inclusive("@").enumerate() {
+                    if pi == 0 {
+                        continue;
+                    }
+
+                    if !part.starts_with(|s: char| s.is_ascii_alphanumeric()) {
+                        continue;
+                    }
+
+                    let Some(username) = part.split_whitespace().next() else {
+                        continue;
+                    };
+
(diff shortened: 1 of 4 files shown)


rustsec/src/advisory/linter.rs
```diff
@@ -61,6 +64,45 @@ impl Linter {
             errors: vec![],
         };
 
+        // Lint against user mentions to avoid spamming
+        if advisory_parts.markdown.contains(" @") {
```
taiki-e: I think we also need to check `(@` ([example](https://github.com/rustsec/advisory-db/blob/5e319f89424272fa0cad22463f444ab44c93df7d/crates/smallvec/RUSTSEC-2021-0003.md?plain=1#L28)) (and maybe `\n@` too). (In https://github.com/rustsec/advisory-db/issues/3046, I used ``[^\[`\w.\/]@\w+`` to search.)

taiki-e (commented):
Thanks!

> Do you think the last commit here would be sufficient?

Yes, detecting and rejecting such cases using lint seems to be the best approach here.

> Have you tested that the explicit links don't trigger notifications?

Yes. Renovate uses this format to avoid mention. For example, https://github.com/lukethacoder/tauri-solid-example/pull/79 which including the following markdown didn't notify me.

```md
thanks [@​taiki-e](https://redirect.github.com/taiki-e))
```

rustsec/src/advisory/linter.rs
djc: Revised the code to be more flexible.

djc:
The audit job issue is resolved in

- https://github.com/rustsec/rustsec/pull/1629

djc:
@tarcieri @LawnGnome friendly ping?

LawnGnome (approved):
Sorry; fell through the notification gaps. LGTM!

PR #1634, opened 2026-07-10
Lints

### LINKED ISSUE #1297: cargo-audit from git fails to build on rustc 1.83 due to `#![deny(warnings)]`
Observed on commit c460e54ecd66a5da8d9725ee419b5199b552e0e6 on rustc 1.83 stable:

```
error: non-local `impl` definition, `impl` blocks should be written at the same level as their item
  --> admin/src/commands.rs:21:10
   |
21 | #[derive(Command, Debug, Parser, Runnable)]
   |          ^------
   |          |
   |          `Command` is not local
   |          move the `impl` block outside of this constant `_DERIVE_Command_FOR_AdminSubCmd`
22 | pub enum AdminSubCmd {
   |          ----------- `AdminSubCmd` is not local
   |
   = note: the derive macro `Command` defines the non-local `impl`, and may need to be changed
   = note: the derive macro `Command` may come from an old version of the `abscissa_derive` crate, try updating your dependency with `cargo update -p abscissa_derive`
   = note: an `impl` is never scoped, even when it is nested inside an item, as it may impact type checking outside of that item, which can be the case if neither the trait or the self type are at the same nesting level as the `impl`
   = note: items in an anonymous const item (`const _: () = { ... }`) are treated as in the same scope as the anonymous const's declaration for the purpose of this lint
note: the lint level is defined here
  --> admin/src/lib.rs:5:9
   |
5  | #![deny(warnings, missing_docs, trivial_casts, unused_qualifications)]
   |         ^^^^^^^^
   = note: `#[deny(non_local_definitions)]` implied by `#[deny(warnings)]`
   = note: this error originates in the derive macro `Command` (in Nightly builds, run with -Z macro-backtrace for more info)
```

`cargo update` didn't help.

`cargo install cargo-audit` (from crates.io) still works, I hope it ignores `#[deny(warnings)]`

djc:
Based on:

- https://github.com/rustsec/rustsec/pull/1632

Fixes #1297.


Shnatsel (approved):
The basic direction looks good, that `deny` is a major footgun and I expected it to cause problems along these lines.

We already have `-D warnings` on CI so warning in the code is enough.

I haven't reviewed closely but the changes are mostly mechanical so I trust you on this.

djc:
Thanks for the review!

PR #1635, opened 2026-07-10
Add Debian Rust data interchange

kpcyrd:
This extends the "Data Interchange" list with Debian, since RustSec's advisory-db is the primary data source for rustsec.debian.net.

image

I'm unsure if they should be displayed all in one row, or if 2 per row is better, I also took the Debian SVG from the [Debian Logos](https://www.debian.org/logos/index.de.html) page and edited the size ratio to match that of the Github SVG. The Debian logo appears smaller than the others, alternatively I also experimented with a `logo-image-square` css class that sets `max-height: 50px`.


djc (approved):
LGTM, thanks!

PR #1647, opened 2026-07-22
admin: parallel crates.io metadata for synchronizer

djc:
While working on

- #1646 

I noticed the synchronizer runs could take quite long. This PR applies the same strategy as used in

- #1621 

after substantially simplifying the synchronizing code and making error messages more precise.

I've validated that the output is the same before and after this branch.

Best reviewed per commit.


LawnGnome (approved):
There's quite a bit here, but it's mostly pretty mechanical in the commit-by-commit, so I think this is good. Nice work!

PR #1650, opened 2026-07-22
cargo-lock: replace gumdrop with clap

djc:
gumdrop was last released 4 years ago, and we're already using clap for the other stuff.

(With help from Claude.)

### DIFF
--- a/cargo-lock/src/main.rs
+++ b/cargo-lock/src/main.rs
@@ -9,7 +9,7 @@ use cargo_lock::{
     dependency::graph::EdgeDirection,
     package::{self},
 };
-use gumdrop::Options;
+use clap::Parser;
 use petgraph::graph::NodeIndex;
 use std::{
     env, fs, io,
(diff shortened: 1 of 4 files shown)


djc:
Not worried about the semver breakage from the removed implicit `gumdrop` Cargo feature, since I assume most people use `cli` anyway.

(To the extent that people use this crate as a binary? Not sure if that's worth keeping.)

cargo-lock/src/main.rs
```diff
@@ -125,34 +123,21 @@ impl TranslateCmd {
...
     exact: bool,
 
-    // Show inverse dependencies rather than forward dependencies
-    #[options(
-        short = "i",
-        long = "invert",
-        help = "show inverse dependencies _on_ a package, rather than forward dependencies _of_ a package"
-    )]
+    /// Show inverse dependencies _on_ a package, rather than forward dependencies _of_ a package
+    #[arg(short = 'i', long = "invert")]
```
LawnGnome: Not that it really matters, but this could be:

```suggestion
    #[arg(short, long = "invert")]
```

LawnGnome (approved):
I don't know gumdrop at all, so this is more of a +0.5 than a +1, but it seems like the semantics are preserved here.

I think dropping the feature is fine.

cargo-lock/src/main.rs
djc: Nice catch; fixed, thanks for the review!

PR #1651, opened 2026-07-23
admin: embed static assets in binary (again)

djc:
So my earlier PR was wrong:

- https://github.com/rustsec/rustsec/pull/1639

When deploying the site, the publish-web workflow in the advisory-db repo generates the site. It usually works without a full checkout, instead using `cargo install` to install the `rustsec-admin` binary directly from a commit in this repo.

As such, embed the assets in the binary again, but do it without any dependencies. Adds a test to make sure that all the files in `web/static` are in the binary, with the filenames mapped to the correct contents.

### DIFF
--- a/admin/src/web.rs
+++ b/admin/src/web.rs
@@ -545,22 +545,18 @@ fn render_feed(output_path: &Path, advisories: &[AdvisoryData]) {
     feed.write_to(file).unwrap();
 }
 
+/// Copies all the static assets to the output folder
+///
+/// For production, we build the site in the advisory-db publish-web GitHub Actions workflow.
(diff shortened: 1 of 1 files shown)


tarcieri:
Hmm, what's with that semver check? Looks like a false positive?

djc:
> Hmm, what's with that semver check? Looks like a false positive?

See https://github.com/rustsec/rustsec/pull/1650#issuecomment-5048834909:

> Not worried about the semver breakage from the removed implicit gumdrop Cargo feature, since I assume most people use `cli` anyway.

alex (approved):

djc:
@alex do you want to bump the admin commit in the advisory-db repo?

PR #1652, opened 2026-07-23
rustsec: use String to represent advisory arch and os

djc:
As suggested in

- #1627


Shnatsel (approved):
I believe the alternative is `Vec<Arch>` that filters out the unknown ones on loading as opposed to preserving all in `Vec<String>`. And the string approach seems better to me. LGTM.

PR #1660, opened 2026-07-30
cvss: align EPSILON with its documented value

gronke:
The comment transcribes Red Hat's Python rounding with EPSILON = 10**-6, but the constant read 10e-6, which is 1e-5. No observable change: the full Red Hat corpora (vectors_base4, vectors_modified4, vectors_random4 — 578,224 vectors) score identically under either value, since interpolated scores sit far closer to the ties than either epsilon.

### DIFF
--- a/cvss/src/v4/score.rs
+++ b/cvss/src/v4/score.rs
@@ -153,7 +153,7 @@ impl Score {
     #[cfg(feature = "std")]
     pub(crate) fn round_v4(value: f64) -> f64 {
         let value = f64::clamp(value, 0.0, 10.0);
-        const EPSILON: f64 = 10e-6;
+        const EPSILON: f64 = 1e-6;
         ((value + EPSILON) * 10.).round() / 10.
     }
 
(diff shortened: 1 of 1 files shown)


djc:
Hmm, are you sure?

```
Python 3.9.6 (default, Dec  2 2025, 07:27:58) 
[Clang 17.0.0 (clang-1700.6.3.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 10 ** -6
1e-06
```

gronke:
It seems it's the two notations colliding.

```python
% python3
Python 3.14.6 (main, Jun 10 2026, 10:03:53) [Clang 21.0.0 (clang-2100.0.123.102)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 10 ** -6
1e-06
>>> 10e-6
1e-05
>>> (10e-6) == (10 ** -6)
False
```

Python pads the output `1e-06` on output, so that these are equal:

```python
>>> 1e-06 == 1e-6
True
```

Python's `10 ** -6` is 10⁻⁶, which is what the comment documents.
The constant, though, was the float literal `10e-6`, and e-notation means 10 × 10⁻⁶ = 10⁻⁵.

```rust
// The crate's rounding, with the epsilon as a parameter.
fn round_v4(value: f64, epsilon: f64) -> f64 {
    let value = f64::clamp(value, 0.0, 10.0);
    ((value + epsilon) * 10.).round() / 10.
}

fn main() {
    println!("10e-6 = {:e}", 10e-6);
    println!("1e-6  = {:e}", 1e-6);
    println!("10e-6 == 1e-5: {}", 10e-6 == 1e-5);

    // A score 5e-6 below the 8.55 tie: the documented 10^-6 epsilon must leave it below.
    let score = 8.549_995;
    println!("round_v4({score}, 1e-6)  = {}", round_v4(score, 1e-6));
    println!("round_v4({score}, 10e-6) = {}", round_v4(score, 10e-6));
}
```

```
10e-6 = 1e-5
1e-6  = 1e-6
10e-6 == 1e-5: true
round_v4(8.549995, 1e-6)  = 8.5
round_v4(8.549995, 10e-6) = 8.6
```

djc (approved):

djc:
Thanks for the explanation!

PR #1662, opened 2026-07-31
cvss: round v2 ties half up and floor published scores at zero

gronke:
cvss2.py computes in exact decimal arithmetic and rounds `ROUND_HALF_UP`; the crate computes in f64, where products like `9.0 * 0.95` sit one ULP below the tie and round down.
Corpus impact: 856 temporal/environmental values (837 a tenth low, 19 at -0.2); base scores are unaffected.

- `Score::roundup` lifts ties over the boundary before rounding, the same epsilon approach the official CVSS v4 reference calculator takes (`EPSILON = Math.pow(10, -6)` in cvss40.js); closes 833 of the 856. cvss2.py itself needs no epsilon, its decimals are exact.
- The three published scores (base, temporal, environmental) floor at zero; the adjusted-temporal intermediate inside the environmental equation stays unclamped, like the reference (the remaining 23).
- Unit tests pin the epsilon magnitude from both sides (a tie one ULP below 8.55 lifts, a value 5e-6 below stays down) and the floor-vs-intermediate distinction (3.9, not 4.0).

### DIFF
--- a/cvss/src/v2/vector.rs
+++ b/cvss/src/v2/vector.rs
@@ -401,6 +401,34 @@ mod tests {
         (a - b).abs() < 0.0001
     }
 
+    /// The reference implementations compute in exact decimal arithmetic and
+    /// round ties half up; `9.0 * 0.95` must round to 8.6, not stay at the
+    /// `f64` product one ULP below the tie.
+    #[test]
+    fn temporal_tie_rounds_half_up() {
+        let v = "AV:N/AC:L/Au:S/C:C/I:C/A:C/E:F/RL:U/RC:C";
+        let cvss = Vector::from_str(v).expect("parse vector");
+        assert!(approx_eq(cvss.score().value(), 9.0));
+        assert!(
+            approx_eq(cvss.temporal_score().value(), 8.6),
+            "temporal expected 8.6, got {}",
+            cvss.temporal_score().value()
+        );
+    }
+
+    /// The environmental equation hits the same ties through its collateral
+    /// damage and target distribution factors.
+    #[test]
+    fn environmental_tie_rounds_half_up() {
+        let v = "AV:N/AC:M/Au:N/C:C/I:C/A:P/RL:TF/RC:UC/CDP:MH/TD:M/IR:L/AR:H";
+        let cvss = Vector::from_str(v).expect("parse vector");
+        assert!(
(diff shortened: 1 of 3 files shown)


cvss/src/v2/vector.rs
```diff
@@ -82,7 +82,7 @@ impl Vector {
     #[cfg(feature = "std")]
     #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
     pub fn score(&self) -> Score {
-        self.base_score_internal(self.impact())
+        Score::new(self.base_score_internal(self.impact()).value().max(0.0))
```
djc: Nit: I generally find the method version of `max()` a little confusing -- prefer you write these as `Ord::max(_, _)`.

djc (approved):
Thanks!

(Please squash changes addressing review feedback back into the originating commit.)

djc:
Are we not running against Red Hat's test vectors here, or are these cases not covered by those test vectors?

gronke:
> Are we not running against Red Hat's test vectors here, or are these cases not covered by those test vectors?

Both, in effect. The harness runs Red Hat's vectors, but it compares only the base score of each triple, and it skips every vector string longer than 44 characters ([`redhat-v2.rs#L22-L25`](https://github.com/rustsec/rustsec/blob/e956a79146dcc607b031275a3a21272d5b016bc2/cvss/tests/redhat-v2.rs#L22-L25)). In vectors_random2 that skips 99,358 of the 100,000 vectors; of the 642 that remain, 509 carry temporal and 621 carry environmental metrics, yet only their base scores are compared, and vectors_simple2 is base-only by construction. All 856 divergent values are temporal or environmental and every base score agrees, so the suite stayed green.

I am already working on a follow-up branch that extends the harness to assert all three scores of every (base, temporal, environmental) triple and adds the two corpus files it never read (vectors_calculator2, vectors_cvsslib2); that is how these surfaced. Happy to fold that harness change into this PR if you prefer the proof in the same change.

djc:
> I am already working on a follow-up branch that extends the harness to assert all three scores of every (base, temporal, environmental) triple and adds the two corpus files it never read (vectors_calculator2, vectors_cvsslib2); that is how these surfaced. Happy to fold that harness change into this PR if you prefer the proof in the same change.

Sounds good -- happy to review that as followup!

PR #1472, opened 2025-11-29
gate the v3/v4 versions of MetricType

alexanderkjall:
fixes #1471

PR #1416, opened 2025-09-12
[cvss][v4] Fix parsing `Not Defined (X)` for `Confidentiality Requirements (CR)` and `Availability Requirements (AR)` in `Environmental (Security Requirements)`

adrianramos-stratio:
Fixes #1415

PR #1663, opened 2026-07-31
cvss: v2 docs and changelog catch-up

gronke:
lib.rs still carried the CVSS v2.0 TODO and its usage section skipped the v2 module; the changelog stopped at 2.1.0.
Backfills 2.1.1 (#1400), 2.2.0 (#1416, #1421), and the unreleased 3.0.0 changes (#1465, #1472), reconstructed from the tagged releases (their own changelogs also stop at 2.1.0) with crates.io publish dates, and documents the hand-constructed-vector contract on the two v2 scoring helpers.
If you prefer a different backfill shape, happy to adjust.

### DIFF
--- a/cvss/CHANGELOG.md
+++ b/cvss/CHANGELOG.md
@@ -1,15 +1,9 @@
 # Changelog
-All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-
-## Unreleased
-### Fixed
-- v2 temporal and environmental ties round half up, matching the reference ([#1662])
-- v2 published scores floor at zero ([#1662])
-
-[#1662]: https://github.com/rustsec/rustsec/pull/1662
+This file is deprecated.
+Release notes for `cvss` v2.1.1 and later are published on
+[GitHub Releases](https://github.com/rustsec/rustsec/releases?q=cvss&expanded=true).
+The entries below predate the move.
 
 ## 2.1.0 (2025-06-06)
 ### Added
--- a/cvss/src/lib.rs
+++ b/cvss/src/lib.rs
@@ -10,6 +10,10 @@
 //! The [`Cvss`] type provides a unified interface for working with CVSS
 //! vectors.
 //!
+//! The [`v2::Vector`] type provides a fully-featured implementation of CVSS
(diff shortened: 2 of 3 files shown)


cvss/CHANGELOG.md
```diff
@@ -5,12 +5,36 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
```
djc: I don't want to maintain this `CHANGELOG` anymore. In a separate commit, can you replace anything that is covered by GitHub Releases with a note that says this file is deprecated, and links out to GitHub Releases?

gronke: Sure, should I (or you) do that in a separate PR and I'll remove changes to CHANGELOG.md here after all?

**Off-topic**: Just a couple of days ago, streamlined my release flows into the exact opposite. PRs add lines to CHANGELOG.md under `[UNRELEASED]`, and then when cutting a release I convert CHANGELOG.md bullets into plaintext and cut a release from it. The final reason why I decided this way around was the commit history that includes the changelog and even when merging a bunch of PRs last minute, there is no chance extra chance to screw up the release message (which would be annoying for immutable releases on GitHub). I'm happy to spin up an example release flow on my fork that uses https://github.com/gronke/rust-ci#changelog in case you're curious to see it in action.

gronke: The changelog is marked as deprecated now.

djc (approved):

djc:
LGTM, thanks!

cvss/CHANGELOG.md
djc: I like to rely on GitHub's auto-generated release notes since that saves time on chores that IMO have relatively little value, and I've found that having changelogs go through reviews slows things down for little benefit.

(Note that even with immutable releases, the release notes can still be modified -- only the assets and the tag itself are immutable AIUI.)

gronke: > I like to rely on GitHub's auto-generated release notes since that saves time on chores that IMO have relatively little value, and I've found that having changelogs go through reviews slows things down for little benefit.

Good points. You maintain much larger crates, so thanks for sharing your experience.

> Note that even with immutable releases, the release notes can still be modified

This goes well with not having a CHANGELOG.md in the repo, I see. TIL

PR #1664, opened 2026-07-31
cvss: check all three v2 scores across all Red Hat corpora

gronke:
The harness ran Red Hat's vectors but compared only the base score of each triple, and it skipped every vector string longer than 44 characters; in vectors_random2 that skips 99,358 of the 100,000 vectors, and the 642 short vectors that remain carry temporal metrics that were never asserted. Now every (base, temporal, environmental) triple is asserted in full (None entries, vectors without metrics of the group, assert nothing), and the two corpus files the harness never read join: vectors_calculator2 and vectors_cvsslib2.

The v2 guide's third worked example (CVE-2003-0062, the AV:L/AC:H corner) completes the Section 3.3 set.
This is the harness through which the #1662 divergences surfaced.

### DIFF
--- a/cvss/src/v2/vector.rs
+++ b/cvss/src/v2/vector.rs
@@ -561,4 +561,45 @@ mod tests {
             environmental
         );
     }
+
+    #[test]
+    fn spec_cve_2003_0062() {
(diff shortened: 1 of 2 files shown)


djc (approved):
LGTM, thanks!

Issue #376, opened 2021-05-26
Json output does not respect deny options


alan-signal:
0.14.1

# Reproduction steps

1. Create a repo with a warning, e.g.:
```
[dependencies]
cpuid-bool = "0.2"
```
2. Run `$cargo audit` and see the warning.
3. Promote that warning to an error and see the error `$cargo audit --deny warnings`.

4. Now run with and without `--deny warnings` and `--json`.

`$ cargo audit --json`
`$ cargo audit --deny warnings --json`

Expect
==

The warning to be promoted to an error in the json.

Actual
==
No difference is seen, it remains a warning:

```
  "warnings": {
    "unmaintained": [
      {
        "kind": "unmaintained",
        "package": {
          "name": "cpuid-bool",
          "version": "0.2.0",
      ...
      }
    ]
  }
```

Additionally, you can reproduce this with a `.cargo/audit.toml`:

```toml
 [output]
 deny = ["unmaintained", "unsound", "yanked"]
 quiet = false
```

Consequences
==

The [audit-check](https://github.com/actions-rs/audit-check) which uses `--json` cannot be made to fail on warnings. See: https://github.com/actions-rs/audit-check/issues/132

Ekleog-NEAR:
I’m hitting this too, it seems like it’s not possible to have the `audit-check` action deny warnings currently, unfortunately. I guess I’ll stay with a hand-rolled audit action, even though it’s sad considering the work that seems to have been put into the audit-check action :)

Shnatsel:
In the latest release of `rustsec` I have aligned warnings to contain all the information that errors do, so this should not be too hard to fix now.

I'm not sure I will be able to get around to fixing this myself soon, though. A PR would be welcome.

PR #1671, opened 2026-08-07
Report in JSON which warnings --deny promotes to errors

Eljees:
Refs #376. @Shnatsel — you said [a PR would be welcome](https://github.com/rustsec/rustsec/issues/376#issuecomment-1818119364); this is one shape of it, and I'd rather check the shape with you than assume.

First, what current `main` actually does. The MCVE from the issue (`cpuid-bool = "0.2"`):

| run | exit | JSON body |
| --- | --- | --- |
| `cargo audit --json` | 0 | `vulnerabilities.found: false`, `warnings.unmaintained: 1` |
| `cargo audit --deny warnings --json` | **1** | **byte-identical** |

So the *exit code* is already correct — `should_exit_with_failure` counts denied warnings regardless of output format. What's missing is that the JSON says nothing about it. The terminal output distinguishes the two cases (`error: 1 denied warning found!` versus `warning: 1 allowed warning found`); the JSON doesn't, so a consumer that parses it — [audit-check](https://github.com/actions-rs/audit-check), which is the consequence the issue names — can't tell a denied warning from an allowed one.

This carries that same fact across. `warning_summary` is added next to the existing keys:

```json
"warning_summary": { "denied": 1, "allowed": 0, "deny": ["warnings", "unmaintained", "unsound", "yanked"] }
```

with `"denied": 0, "allowed": 1, "deny": []` when nothing is denied. The numbers come from the existing `count_warnings`, so the JSON and the terminal summary line can't drift apart. The `deny` list is there because the issue asks specifically for the `.cargo/audit.toml` route to work, and a consumer may want to know *why* something was denied — that case is covered and produces `deny: ["unmaintained", "unsound", "yanked"]`.

The report is `#[serde(flatten)]`-ed into the wrapper, so **every pre-existing key stays exactly where it was** — verified by comparing the two JSON bodies key by key. Terminal and SARIF output are untouched.

**On the shape.** The issue says it expects the warning "to be promoted to an error in the json", i.e. moved into `vulnerabilities`. I didn't do that, because `Warning` and `Vulnerability` are different types and moving denied warnings would change what every existing `--json` consumer sees, including ones that don't set `deny` at all. An additive key can't break anyone. If you'd rather have the promotion — or a per-warning `"denied": true` instead of a summary — say which and I'll redo it; the change is small either way.

`cargo build`, `cargo fmt --check`, `cargo clippy` (0 warnings, same as the clean tree) and `cargo test -p cargo-audit` are all green.

AI-assisted (LLM used for drafting); the runs above are mine.

### DIFF
--- a/cargo-audit/src/presenter.rs
+++ b/cargo-audit/src/presenter.rs
@@ -18,6 +18,7 @@ use rustsec::{
 };
 #[cfg(feature = "binary-scanning")]
 use rustsec::{advisory::affected::FunctionPath, binary_scanning::BinaryReport};
+use serde::Serialize;
 
 #[cfg(feature = "binary-scanning")]
 use crate::binary_scanning::SymbolSet;
@@ -42,6 +43,38 @@ pub struct Presenter {
     /// Binary contents for affected-function analysis
     #[cfg(feature = "binary-scanning")]
     binary_contents: Option<Vec<u8>>,
+}
+
+/// A report as it is written in JSON.
+///
+/// The report itself is flattened in, so every key an existing consumer reads
+/// stays exactly where it was; `warning_summary` is added alongside it.
+#[derive(Serialize)]
+struct JsonReport<'a> {
+    #[serde(flatten)]
+    report: &'a rustsec::Report,
+
+    /// What `--deny` made of the warnings above.
+    warning_summary: WarningSummary<'a>,
+}
+
+/// How the configured `deny` options apply to the warnings in a report.
+///
+/// The terminal output already says whether a warning was denied or allowed
+/// -- "1 denied warning found!" versus "1 allowed warning found" -- and the
+/// process exits non-zero for a denied one. This carries the same fact into
+/// the JSON, which otherwise comes out byte-identical whether `--deny` was
+/// given or not.
+#[derive(Serialize)]
+struct WarningSummary<'a> {
+    /// Number of warnings the `deny` configuration promotes to errors.
+    denied: u64,
+
+    /// Number of warnings it leaves as warnings.
+    allowed: u64,
+
+    /// The `deny` options in effect, from `--deny` or `[output] deny`.
+    deny: &'a [DenyOption],
 }
 
 impl Presenter {
@@ -119,8 +152,18 @@ impl Presenter {
     ) {
         match self.config.format {
             OutputFormat::Json => {
+                let (denied, allowed) = self.count_warnings(report);
+                let json_report = JsonReport {
+                    report,
+                    warning_summary: WarningSummary {
+                        denied,
+                        allowed,
+                        deny: &self.config.deny,
+                    },
+                };
+
                 let mut stdout = io::stdout().lock();
-                serde_json::to_writer(&mut stdout, &report).unwrap();
+                serde_json::to_writer(&mut stdout, &json_report).unwrap();
                 // End with a newline as a terminator/separator. Another json report may follow.
                 writeln!(&mut stdout).unwrap();
                 return;


djc:
Please don't let your LLM write a PR description for you.

It looks like this is wrapping the actual report in a weird different layer which looks like it might be incompatible or would make it easier to make it incompatible with existing consumers in the future.

It would be helpful if you can explain concisely (that is, in your own words) what the proposed solution is.

Eljees:
Fair on the description — it was too long and it wasn't in my voice. I use an LLM to draft; the analysis and the runs behind it are mine, and I should have cut it down before posting. Sorry. Short version:

The exit code is already right today. The JSON isn't: `cargo audit --json` and `cargo audit --deny warnings --json` produce byte-identical output, so anything parsing the JSON cannot tell a denied warning from an allowed one. That is the gap #376 describes.

The patch adds one key next to the existing ones:

```json
"warning_summary": { "denied": 1, "allowed": 0, "deny": ["unmaintained"] }
```

There is no extra layer around the report in the output. `#[serde(flatten)]` is on the report field, so every existing key keeps its name and its position at the top level and nothing is nested or renamed; I diffed the JSON before and after, key by key.

The one real consequence of `flatten` is that serde serialises through a map instead of a struct. For `serde_json` the bytes are the same, but if you'd rather not take that on, the alternative is to put the two counts on `rustsec::Report` itself and drop the wrapper struct entirely. I'm happy to redo it that way — say the word.

I also deliberately did not do what the issue literally asks, which is to move denied warnings into `vulnerabilities`: that changes what every existing `--json` consumer sees, including ones that never set `deny`. If you'd prefer that, or a per-warning `"denied": true` flag instead of a summary, tell me which and I'll rewrite.

djc:
> I also deliberately did not do what the issue literally asks, which is to move denied warnings into vulnerabilities: that changes what every existing --json consumer sees, including ones that never set deny. If you'd prefer that, or a per-warning "denied": true flag instead of a summary, tell me which and I'll rewrite.

Can you do some analysis on these options to discuss the different trade-offs and to be made?

IMO moving warnings into `denied` seems the most intuitive behavior.

Eljees:
Three options. What decides between them is what a `--json` consumer that never passes `--deny` sees.

**A — summary key (what this PR does now).** `"warning_summary": {"denied": N, "allowed": N, "deny": [...]}` next to the existing keys. Nothing existing moves or is renamed, so a run without `--deny` produces the same bytes as today. A consumer that wants per-warning denial has to derive it by intersecting `deny` with the keys of `warnings` — possible, but it is work I am pushing onto them. The `#[serde(flatten)]` wrapper is not essential to this option: I can put the fields on `Report` directly and delete the wrapper struct.

**B — per-warning `"denied": true`.** No wrapper, no derivation, the bit sits on the item it describes. Compatibility is the same as A, except it breaks anything deserialising warnings with `deny_unknown_fields`. The weakness is the empty case: with zero warnings the output says nothing about what `--deny` was set to, so a CI job cannot tell "deny configured, nothing denied" from "deny not configured".

**C — move denied warnings into `vulnerabilities`.** I agree this is the most intuitive reading, and it only changes output for runs that already exit non-zero. But the types do not allow it for every kind as they stand:

- `Vulnerability` (`rustsec/src/vulnerability.rs`) has `advisory: advisory::Metadata` and `versions: advisory::Versions`, both non-optional.
- `Warning` (`rustsec/src/warning.rs`) has `advisory`, `affected` and `versions` all as `Option`, and yanked warnings are constructed as `Warning::new(WarningKind::Yanked, pkg, None, None, None)` in `cargo-audit/src/auditor.rs`.

So `--deny yanked` has nothing to build a `Vulnerability` from. That leaves three ways out: leave yanked behind and move only the advisory-backed kinds; make those two fields optional on `Vulnerability`, which is a breaking change for every consumer rather than only for the ones setting `--deny`; or synthesise an advisory, which I would rather not do.

The second cost of C is that the kind — `unmaintained` vs `unsound` vs `yanked` — is not recoverable once the entry sits in `vulnerabilities`, unless it also stays in `warnings`, and then `vulnerabilities.count` plus the warning map double-count it.

My read: C for the advisory-backed kinds plus B's flag for yanked is coherent, but it is two shapes for one concept. B alone is one shape that covers all four kinds. A is the least invasive and the least useful of the three.

If you are fine with yanked staying a warning, I will write C. Otherwise say B and I will write that instead.

Eljees:
@djc gentle ping — the trade-off analysis you asked for is in my comment above: A (summary key, what the PR does now), B (per-warning `"denied": true`), C (move denied warnings into `vulnerabilities`). C is the one you called most intuitive and I don't disagree, but it needs the type changes I listed there, so I'd rather have your pick before rewriting. Say which one and I'll redo the PR to it.

Eljees:
@djc — a week on, so just the open question rather than more analysis.

It reduces to one thing: whether you are fine with `--deny yanked` staying a warning. `Warning` carries `advisory`, `affected` and `versions` as `Option` and yanked warnings are built with all three `None`, so there is nothing to construct a `Vulnerability` from. C therefore can't cover all four kinds without either making those two fields optional on `Vulnerability` — which breaks every consumer, not only the ones passing `--deny` — or synthesising an advisory, which I'd rather not do.

If yanked staying a warning is acceptable, I'll write C. If not, say B and I'll write that instead. Either is fine by me; I just don't want to guess and hand you the wrong shape.

djc:
> **B — per-warning `"denied": true`.** No wrapper, no derivation, the bit sits on the item it describes. Compatibility is the same as A, except it breaks anything deserialising warnings with `deny_unknown_fields`. The weakness is the empty case: with zero warnings the output says nothing about what `--deny` was set to, so a CI job cannot tell "deny configured, nothing denied" from "deny not configured".

This sounds like the best option to me. I don't much care about anyone deserializing warnings with `deny_unknown_fields`. Do you care about the empty case? It seems like an unimported edge case to me -- the caller should already know what they set `--deny` to anyway.

Eljees:
Agreed on B then — and no, I don't care about the empty case either. You're right that the caller already knows what `--deny` was set to; I was optimizing for a consumer that only looks at the warning list in isolation, and that's not realistic. Dropping the empty-case requirement. Will push B as-is.

PR #1672, opened 2026-08-09
rustsec: clean untracked files out of repository workdir

djc:
To fix problems like

- https://github.com/rustsec/advisory-db/issues/3131

in the future.

### DIFF
--- a/rustsec/src/repository/git/commit.rs
+++ b/rustsec/src/repository/git/commit.rs
@@ -1,5 +1,14 @@
 //! Commits to the advisory DB git repository
 
+use std::{
+    collections::HashSet,
+    fs,
+    path::Path,
+    time::{Duration, SystemTime},
+};
+
+use gix::bstr::BString;
+
 use crate::{
     error::{Error, ErrorKind},
     repository::{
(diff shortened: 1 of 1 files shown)


alex:
This seems like it does what it says, but I'm not sure I understand what's required to reproduce the issue.

Is it simply "have a checkout at revision X, we remove an adivsory, update"? That would quite surprirse me if it wa sso brittle!

djc:
> This seems like it does what it says, but I'm not sure I understand what's required to reproduce the issue.
> 
> Is it simply "have a checkout at revision X, we remove an adivsory, update"? That would quite surprirse me if it wa sso brittle!

The issue people are reporting is because we moved an advisory, which leads to there being two advisories with the same file name for repositories that previously checked out the older commit. In most cases I guess we simply don't notice when there are stale files.

IMO this is arguably a bug in gix where their `checkout()` in an existing workdir doesn't unlink files that have become untracked, unlike CLI git.

alex:
Yes, that does sound like a bug in gix.

djc:
Submitted PR upstream:

- https://github.com/GitoxideLabs/gitoxide/pull/2900

PR #1677, opened 2026-08-12
rustsec: reduce error API surface

djc:
I think this API surface doesn't need to exist, and we can rely on `Error` and `Display` impls.


djc:
@tarcieri @LawnGnome ping?

tarcieri (approved):

PR #1682, opened 2026-08-26
cargo-audit: make advisory DB lock timeout configurable

### LINKED ISSUE #974: Make lock timeout configurable
`cargo audit` currently hardcodes how long it will wait to acquire a filesystem lock: 
https://github.com/rustsec/rustsec/blob/7c05b2ec0104a1a7c38322d90baf7627add73a11/cargo-audit/src/auditor.rs#L17-L18

It would be nice to make it configurable through the configuration file and/or the command-line arguments.

ChrisJr404:
Fixes #974.

The wait for the advisory database directory lock was hardcoded to 5 minutes in the auditor. This adds a `database.lock_timeout` config option (in seconds) plus a matching `--db-lock-timeout` CLI flag, wired through `override_config` the same way as the other database settings. When neither is set it falls back to the existing 5 minute default, so behavior is unchanged unless you opt in.

Added a unit test alongside the existing fetch-override one, and documented the option in `audit.toml.example`.


djc:
What is your motivation for this change? Maybe 300s is just a bad default and it should be shorter?

ChrisJr404:
The motivation is CI and shared-cache setups where several `cargo audit` invocations can run against the same advisory DB at once and contend on the lock. In those cases 300s is sometimes too long (a stuck holder blocks the whole job well past a sensible CI budget) and occasionally too short (a cold clone over a slow proxy). Rather than guess a single better number, exposing it lets fast-fail CI drop it to a few seconds while slow environments raise it.

That said, I would not object to also lowering the default if you think 300s is simply too generous; the two changes are complementary. Happy to go either way.

cargo-audit/src/commands/audit.rs
```diff
@@ -130,6 +130,14 @@ pub struct AuditCommand {
     #[arg(long = "stale", help = "allow stale database")]
     stale: bool,
 
+    /// Seconds to wait for the advisory database directory lock before giving up
+    #[arg(
+        long = "db-lock-timeout",
+        value_name = "SECONDS",
+        help = "seconds to wait for the advisory DB directory lock (default: 300)"
+    )]
+    db_lock_timeout: Option<u64>,
```
djc, line 139: Suggest we make this `db_lock_timeout: Duration` directly, setting the default inside the `#[arg()]` attribute.

djc (commented):
Also, please rebase to address the CI failures.

PR #1683, opened 2026-08-27
Avoid panic in rustsec_refs_imported on short/malformed URLs

### LINKED ISSUE #1681: rustsec_refs_imported() could possibly panic on short or malformed rustsec.org URLs
### Summary ###

I see hardcoded byte-range slice [31..48] usage in the function rustsec_refs_imported()

rustsec/src/osv/advisory.rs

```
    /// Try to extract RustSec alias id from OSV advisory metadata
    pub fn rustsec_refs_imported(&self) -> Vec<Id> {
        let mut refs: Vec<Id> = self
            .references
            .iter()
            .filter(|r| {
                r.url
                    .as_str()
                    .starts_with("https://rustsec.org/advisories/")
            })
            .map(|r| Id::from_str(&r.url.as_str()[31..48]).expect("Invalid rustsec url"))
            .collect();
        refs.sort();
        refs.dedup();
        refs
    }
```

It seems that the prefix "https://rustsec.org/advisories/" is 31 bytes.
The slice [31..48] assumes 17 more bytes for the following (the length of RUSTSEC-YYYY-NNNN) ?
URLs that match the prefix but are shorter than 48 bytes cause an out-of-bounds panic. 
Also, URLs that are long enough but don't contain a valid advisory ID cause a panic from expect.

### Reproduce ####

Create a reference URL that starts with https://rustsec.org/advisories/ but is shorter than 48 bytes, then call rustsec_refs_imported()

chiliec:
Fixes #1681.

## Problem

`OsvAdvisory::rustsec_refs_imported` (in `rustsec/src/osv/advisory.rs`) extracted each advisory id with a hardcoded byte slice and an `expect`:

```rust
.filter(|r| r.url.as_str().starts_with("https://rustsec.org/advisories/"))
.map(|r| Id::from_str(&r.url.as_str()[31..48]).expect("Invalid rustsec url"))
```

The prefix is 31 bytes and the slice assumes a further 17 bytes for `RUSTSEC-YYYY-NNNN`. A reference URL that starts with the prefix but is **shorter than 48 bytes** panics with an out-of-bounds slice (`end byte index 48 is out of bounds for string of length 31`); a prefix-matching URL whose id does not parse panics on the `expect`. Reference URLs come from OSV advisory data, so a single malformed entry aborts the whole import.

## Fix

Strip the prefix and the `.html` suffix that the exporter appends (`format!("https://rustsec.org/advisories/{}.html", ...)`), parse the remaining id fallibly, and keep only genuine RustSec ids:

```rust
const PREFIX: &str = "https://rustsec.org/advisories/";
...
.filter_map(|r| {
    let rest = r.url.as_str().strip_prefix(PREFIX)?;
    let id_str = rest.strip_suffix(".html").unwrap_or(rest);
    Id::from_str(id_str).ok().filter(Id::is_rustsec)
})
```

Well-formed URLs yield the same ids as before; short or malformed ones are skipped rather than panicking.

## Tests

Adds a `tests` module with two cases:

- `extracts_ids_from_well_formed_rustsec_urls` — the documented happy path still returns the id.
- `short_or_malformed_rustsec_urls_do_not_panic` — a truncated URL (shorter than the old `[31..48]` slice) and unparseable ids are skipped without panicking.

## Verification (local, Rust 1.98)

- **RED→GREEN**: with the original code restored, `short_or_malformed_rustsec_urls_do_not_panic` fails with `end byte index 48 is out of bounds for string of length 31`; with the fix both tests pass.
- `cargo test -p rustsec --features osv-export --lib osv::` — 24 passed, 0 failed.
- `cargo fmt --check` and `cargo clippy` are clean on the changed file.


gkamathe:
Hello @chiliec , seems we attempted to fix the same issue [1] :) at the exact same time [2], good co-incidence. However your change seems to have tests to verify as well which is great.

[1] https://github.com/rustsec/rustsec/issues/1681
[2] https://github.com/rustsec/rustsec/pull/1684

chiliec:
Heads-up that the red CI here is unrelated to this change (which only touches `rustsec/src/osv/advisory.rs`) — it's a repo-wide dependency issue:

- **`semver`** fails at `cargo update`: `failed to select a version for the requirement bisync = "^0.3.0"` — both matching versions (`0.3.0`, `0.3.1`) are yanked. In fact **every published `bisync` version is now yanked** on crates.io (the author yanked the whole crate on 2026-08-24/26), so nothing satisfies `^0.3.0`.
- **`self-audit`** trips `RUSTSEC-2026-0258` plus a yanked `rustsec-admin` in the lockfile.

Both affect any PR against the repo right now, not this one. This will likely need the `bisync` dependency bumped/replaced (or the crate un-yanked upstream) before CI can go green here. Happy to rebase once that's sorted.

rustsec/src/osv/advisory.rs
```diff
@@ -293,15 +293,22 @@ impl OsvAdvisory {
...
-                    .as_str()
-                    .starts_with("https://rustsec.org/advisories/")
+            .filter_map(|r| {
+                // Strip the known prefix and the human-readable ".html"
+                // suffix that the exporter appends, then parse the remaining
+                // advisory id. A URL that merely starts with the prefix but is
+                // truncated or malformed simply yields no id, rather than
+                // panicking on an out-of-bounds byte slice or an `expect` on an
+                // unparseable value. Only genuine RustSec ids are kept, since
+                // that is what this function is documented to return.
```
djc, line 307: I think we can just cut this comment, it doesn't seem to add much value.

rustsec/src/osv/advisory.rs
```diff
@@ -358,3 +365,55 @@ fn guess_url_kind(url: &Url) -> OsvReferenceKind {
 fn rustsec_date_to_rfc3339(d: &crate::advisory::Date) -> String {
     format!("{}-{:02}-{:02}T12:00:00Z", d.year(), d.month(), d.day())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn advisory_with_reference_urls(urls: &[&str]) -> OsvAdvisory {
```
djc, line 373: Nit: please move the helper below the tests.

rustsec/src/osv/advisory.rs
```diff
@@ -358,3 +365,55 @@ fn guess_url_kind(url: &Url) -> OsvReferenceKind {
...
+        let refs: String = urls
+            .iter()
+            .map(|u| format!("{{\"type\":\"ADVISORY\",\"url\":\"{}\"}}", u))
+            .collect::<Vec<_>>()
+            .join(",");
```
djc, line 378: Can we turbofish `::<String>` instead of using the type annotation? Would be nicer IMO.

djc (commented):
Looks okay to me modulo some nits, thanks!

The audit failure will go away once my PR is merged:

- https://github.com/rustsec/rustsec/pull/1685

djc:
Also, please rebase on `main`? That should take care of the CI failures.

PR #1684, opened 2026-08-27
fix: proper parsing of rustsec advisory url

gkamathe:
### Summary ###

Fix for issue - [1681](https://github.com/rustsec/rustsec/issues/1681)

Current extraction for "RustSec alias id from OSV advisory metadata" has hard-coded slice calculations which has a possibility of panicking if the URL is not as per expectations. This change attempts a PREFIX oriented url parsing change. 

### Verify ###

Ran tests using cargo test, did not see any failures


gkamathe:
Closing this request as there is another previous MR [1] which aims to solve the same issue and has tests included in it.

[1] https://github.com/rustsec/rustsec/pull/1683

Issue #779, opened 2023-01-04
Private registry support - proposal


rbtcollins:
#261 notes that authenticated registries don't work and #604 that sparse registries don't work either.

It seems to me that since the rustsec database is for *public* crates, there is no value in attempting to process - or update - private registries, nor crates from private registries. Cargo itself has already generated the transitive closure of dependencies in building the `Cargo.lock` file itself, so a mode where `Cargo.lock` is presumed fresh, and just examined for crates.io dependencies seems like it will avoid all the issues with Cargo itself being a moving target.

tarcieri:
As both those issues note, interactions with the crates.io index occur via the `crates-index` crate:

https://github.com/frewsxcv/rust-crates-index

So issues regarding authenticated registries or sparse index support should be filed there.

Yes, the interactions are only for checking for yanked crates in the crates.io index (and now that `cargo` has some built-in functionality like this it's of debatable value).

We should probably skip fetching the index if the Cargo.lock being examined has no crates with a `source` of crates.io.

rbtcollins:
I think I didn't make my suggestion clear enough. I'm suggesting that private registries should not be updated or examined, thus avoiding all interactions with the crates.io index *for those dependencies*.  And in fact some code inspection shows that this is in fact the case. See #780 which will fix the sparse issue, but doesn't go as far as skipping analysis of private registry crates.

tarcieri:
@rbtcollins that’s exactly what I was saying here:

> We should probably skip fetching the index if the Cargo.lock being examined has no crates with a source of crates.io.

rbtcollins:
Thats a pretty edge case though - I expect the common case of private registries to be some internal crates, but a vast majority of dependencies coming from crates.io.

tarcieri:
In that case, so long as the yank-check feature is enabled, we still need to fetch the crates.io index.

I'm not sure what else you're expecting?

rbtcollins:
Have a look at my patch please; it solves the problem I was encountering. I still don't understand the question about fetching crates.io index - that hasn't ever been a concern I was raising.

tarcieri:
Ok, you linked to issues regarding fetching the crates index, hence my confusion

PR #1694, opened 2026-09-11
Match advisories against both crates.io source spellings

arpitjain099:
`Query::matches` compares the advisory source to the package source by raw kind and URL:

```rust
// TODO(tarcieri): better source comparison?
if advisory_source.kind() != package_source.kind()
    || advisory_source.url() != package_source.url()
{
    return false;
}
```

An advisory that names no source falls back to `SourceId::default()`, which is `Registry` + `https://github.com/rust-lang/crates.io-index`. A lockfile that reaches crates.io through an alternative-registry entry records `SparseRegistry` + `https://index.crates.io/`. Both halves differ, so every advisory for that package is filtered out.

What makes it quiet is that the package was already admitted. `Database::query_vulnerabilities` decides what to audit with `Source::is_default_registry`, which knows both spellings, so the crate is counted in the "scanning N crate dependencies" banner and then has all of its advisories dropped with no diagnostic. `find_warnings` runs through the same query, so unmaintained and unsound warnings go with them.

Two lockfiles differing in exactly one line, `smallvec 0.6.9`, built with `--no-fetch` against the same advisory-db:

```
                          registry+...crates.io-index   sparse+https://index.crates.io/
cargo-audit (current)                  4 advisories                       0
cargo-audit (this PR)                  4 advisories                       4
```

The advisories missed are RUSTSEC-2021-0003, RUSTSEC-2019-0009 and RUSTSEC-2019-0012, all CVSS 9.8, plus the RUSTSEC-2018-0018 unsound warning. The `registry+` column is the positive control and is unchanged.

To produce such a lockfile, cargo does not write `sparse+` for ordinary dependencies even under the sparse protocol; I checked `CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse` and a `[registries.crates-io]` index override, and both still emit `registry+`. What does produce it is crates.io named as an alternative registry:

```toml
[registries]
criosparse = { index = "sparse+https://index.crates.io/" }
```

which is the shape reported in #868.

The fix reuses `is_default_registry` so the two spellings are one registry here as well. Scope worth being explicit about: an organisation proxying crates.io through its own sparse host is filtered out earlier at `database.rs:152` because that URL is not crates.io, which is the #779 design decision and is deliberately untouched.

Tests in `rustsec/tests/query.rs`: `matches_both_crates_io_spellings` fails on main and passes here, and `does_not_match_a_different_registry` pins that an unrelated sparse registry still does not match, passing either way.


djc:
Please edit the PR description and comments to be more concise and in your own words.

PR #1698, opened 2026-09-15
Take semver-compatible dependency updates

djc:
Including a rustls update for https://rustsec.org/advisories/RUSTSEC-2026-0285.html.


djc:
@tarcieri ping?

tarcieri (approved):

PR #1701, opened 2026-09-18
platforms 4.2.1

djc:


tarcieri:
Hmm, fix the self-audit first so CI is green?

LawnGnome (approved):

